PR #1, opened 2016-03-02
Fix services sles

darix:
Fix numerical check of osrelease
    
    After making the version check numerical in 9975508 it no longer matched
    SLES 11 properly to use the rh_service module as '11.4 > 11' evaluates
    to true. Without using the rh_service module, not all methods are
    implemented to use the service state on sle11.

    Make the suse check consistent with rh_service.py

### DIFF
--- a/salt/modules/rh_service.py
+++ b/salt/modules/rh_service.py
@@ -66,8 +66,8 @@ def __virtual__():
             return (False, 'Cannot load rh_service module: '
                            'osrelease grain, {0}, not a float,'.format(osrelease))
         if __grains__['os'] == 'SUSE':
-            if osrelease > 11:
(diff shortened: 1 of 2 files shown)


isbm:
I don't think this is the place for pull req. Original salt Git repo is.

darix:
we can push from here afterwards. i put the PR up so more people look at it before we push it upstream.

isbm:
This is not correct. All PRs should be upstream from the beginning and discussed there from the day on. He we only taking commits from the upstream and nothing else.

****Please****

isbm:
@darix [JFYI](https://github.com/openSUSE/salt/wiki#policies) Please close here.

kkaempf:
@darix thanks for the PR. I'd like to see a test case, can you please provide one ?

PR #2, opened 2016-04-05
Prevent metadata download when getting installed products

mcalmer:
--disable-repos temporary disable all repos for this call. 

@isbm , @dmacvicar - maybe you want to have a look?

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -1308,7 +1308,10 @@ def list_products(all=False, refresh=False):
 
     ret = list()
     OEM_PATH = "/var/lib/suseRegister/OEM"
-    cmd = _zypper('-x', 'products')
+    cmd = _zypper()
+    if not all:
+        cmd.append('--disable-repos')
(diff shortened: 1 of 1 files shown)


salt/modules/zypper.py
```diff
@@ -1308,7 +1308,10 @@ def list_products(all=False, refresh=False):
 
     ret = list()
     OEM_PATH = "/var/lib/suseRegister/OEM"
-    cmd = _zypper('-x', 'products')
+    cmd = _zypper()
+    if not all:
+        cmd.append('--disable-repos')
+    cmd.extend(['-x', 'products'])
```
dmacvicar: I think `--disable-repos` is a global option, and therefore has to go before `products`

mcalmer: It does.

cmd.append(--disable-repos')
cmd.extend(['-x', 'products'])

dmacvicar: Sorry, you are right.

:+1:

mcalmer:
manually merged.

PR #3, opened 2016-04-15
align OS grains from older SLES with current one (bsc#975757)

mcalmer:
Try to unify the os grains between SLE11 and SLE12 independent of existing os-release file:

This commit change SLES11 SP3 and lower in the following way:

    oscodename: n.a => SUSE Linux Enterprise Server 11 SP3
    osrelease:   "11 SP3" => 11.3
    osrelease_info: ["11 SP3"] => [11, 3]

@isbm , @dmacvicar : this is a proposal. What do you think?

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1185,14 +1185,19 @@ def os_data():
                         for line in fhr:
                             if 'enterprise' in line.lower():
                                 grains['lsb_distrib_id'] = 'SLES'
+                                grains['lsb_distrib_codename'] = re.sub(r'\(.+\)', '', line).strip()
                             elif 'version' in line.lower():
                                 version = re.sub(r'[^0-9]', '', line)
                             elif 'patchlevel' in line.lower():
                                 patch = re.sub(r'[^0-9]', '', line)
                     grains['lsb_distrib_release'] = version
                     if patch:
-                        grains['lsb_distrib_release'] += ' SP' + patch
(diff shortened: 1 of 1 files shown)


isbm:
@mcalmer I would try upstreaming this instead opening PR here. Because if someone else will re-use our minions with another master (or the other way around) we are in a trouble.

mcalmer:
@isbm yes, but first *we* need to agree that these are the values we want to see here.

dmacvicar:
Why would we be "in trouble" if someone uses a supported product/minion with an unsupported master?

This is exactly the kind of bugs we should put in our package and tests first and then upstream, as upstream will assume we know better and accept anything we do regarding our distributions.

LGTM :+1: 

I don't like the fact that we do our own Dockerfiles giving that we have official SLES images. We could have jenkins `unrpm` them from http://nu.novell.com/SUSE/Updates/SLE-Module-Containers/12/x86_64/update/x86_64 and push them to our registry.

mcalmer:
> I don't like the fact that we do our own Dockerfiles giving that we have official SLES images. We could > have jenkins unrpm them from http://nu.novell.com/SUSE/Updates/SLE-Module-Containers/12/x86_64/update/x86_64 and push them to our registry.

AFAIK I use the official SLES images as base, but they are very basic and we need more to test salt.

dmacvicar:
Ok, I just realized the Docker team is also using suma-docker-registry.suse.de and probably pushing the images from the rpms there.

PR #6, opened 2016-05-20
libzypp plugin rework

isbm:


isbm:
I forgot that, better I just close. :laughing:

PR #11, opened 2016-09-29
yum plugin opensuse 2015.8.7

isbm:
## \* WORK IN PROGRESS *

Splits the yum notification plugin from the SUSE Manager


isbm:
Picked up manually. Closing.

PR #14, opened 2016-12-14
Avoid failures on SLES 12 SP2 because of new systemd TaskMax limit (bsc#985112)

moio:
This avoids stalls in case `worker_threads` are set high enough to bump into systemd's `DefaultTaskMax` limit in SLES 12 SP2 (512).

See: https://bugzilla.suse.com/show_bug.cgi?id=985112


dincamihai:
@mcalmer there's not much to review here. I think we can trust @moio's tests

kkaempf:
Shouldn't this be pushed upstream ?

PR #19, opened 2017-01-12
Rpm installdate timezone fix

hustodemon:
Previously datetime.fromtimestamp was used. If used without additional
parameters, this method returns date in the local timezone.

Our code took the result of fromtimestamp, appended 'Z' and returned
this string. This is wrong as 'Z' means UTC (the client code parses this
value as UTC, but it's in fact local time).

Fixed by using utcfromtimestamp.

### DIFF
--- a/salt/modules/rpm.py
+++ b/salt/modules/rpm.py
@@ -538,7 +538,7 @@ def info(*packages, **attr):
             # Convert Unix ticks into ISO time format
             if key in ['build_date', 'install_date']:
                 try:
-                    pkg_data[key] = datetime.datetime.fromtimestamp(int(value)).isoformat() + "Z"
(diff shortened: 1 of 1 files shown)


hustodemon:
@isbm Please take a look.

isbm:
Unit test?

isbm:
@hustodemon [no merge](https://github.com/openSUSE/salt/wiki/Workflow#writing-salt-toaster-tests) unless unit test or Salt Toaster test provided. :smiley:

mcalmer:
Toaster Test here: https://github.com/openSUSE/salt-toaster/pull/32

mcalmer:
Merged manually in openSUSE-2015.8.12 and openSUSE-2016.11.1

PR #22, opened 2017-03-22
Backporting fix of #29422 from openSUSE-2016.3.0

pkazmierczak:
This fixes [bug 1019386](https://bugzilla.suse.com/show_bug.cgi?id=1019386). Original commit message below:

The SSH_PASSWORD_PROMPT_RE regexp used to detect if SSH is requesting a
password can be triggered if the shim is returning data to the server
with text that matches the regex, including inside JSON results. This
then results in the server unable to parse the JSON results. This
patch fixes this issue by looking for the shim delimiter in the output and
disabling the sending of passwords after the delimiter is found.

Fixes #29422.

### DIFF
--- a/salt/client/ssh/shell.py
+++ b/salt/client/ssh/shell.py
@@ -23,6 +23,10 @@ log = logging.getLogger(__name__)
 SSH_PASSWORD_PROMPT_RE = re.compile(r'(?:.*)[Pp]assword(?: for .*)?:', re.M)
 KEY_VALID_RE = re.compile(r'.*\(yes\/no\).*')
 
+# Keep these in sync with ./__init__.py
+RSTR = '_edbc7885e4f9aac9b83b35999b68d015148caf467b78fa39c05f669c0ff89878'
+RSTR_RE = re.compile(r'(?:^|\r?\n)' + RSTR + r'(?:\r?\n|$)')
+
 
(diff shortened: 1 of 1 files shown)


salt/client/ssh/shell.py
```diff
@@ -23,6 +23,10 @@
 SSH_PASSWORD_PROMPT_RE = re.compile(r'(?:.*)[Pp]assword(?: for .*)?:', re.M)
 KEY_VALID_RE = re.compile(r'.*\(yes\/no\).*')
 
+# Keep these in sync with ./__init__.py
+RSTR = '_edbc7885e4f9aac9b83b35999b68d015148caf467b78fa39c05f669c0ff89878'
```
isbm: zomg, this is just a delimeter between STDOUT and STDERR... I recon, "RSTR" stands for "Random String, Truly Real".

isbm:
@pkazmierczak please add also a test for Salt Toaster and reference PR from it here. Thanks!

pkazmierczak:
@isbm done.

salt/client/ssh/shell.py
mcalmer: @isbm this was cherry-picked from upstream. So we should not change it:-)

PR #24, opened 2017-03-31
Add unit test for race condition occurrence

isbm:
On https://github.com/openSUSE/salt/commit/39dee1d1251234ea1882862ea6da8bfda22c6792


dincamihai:
It seems the tests are doing the right thing.
It would be nice to use `Mock` instead of `Makedir` class.

isbm:
@dincamihai done

PR #25, opened 2017-04-07
skip false values from preferred_ip

mjura:
(cherry picked from commit f829d6f9fc4ffa97b2c175a4e915a174631eba72)

### DIFF
--- a/salt/cloud/clouds/dimensiondata.py
+++ b/salt/cloud/clouds/dimensiondata.py
@@ -231,6 +231,8 @@ def create(vm_):
             )
             for private_ip in private:
                 private_ip = preferred_ip(vm_, [private_ip])
+                if private_ip is False:
+                    continue
                 if salt.utils.cloud.is_public_ip(private_ip):
                     log.warning('%s is a public IP', private_ip)
                     data.public_ips.append(private_ip)
--- a/salt/cloud/clouds/nova.py
+++ b/salt/cloud/clouds/nova.py
@@ -886,6 +886,8 @@ def create(vm_):
             )
             for private_ip in private:
                 private_ip = preferred_ip(vm_, [private_ip])
+                if private_ip is False:
+                    continue
                 if salt.utils.cloud.is_public_ip(private_ip):
                     log.warning('{0} is a public IP'.format(private_ip))
                     data.public_ips.append(private_ip)
--- a/salt/cloud/clouds/openstack.py
+++ b/salt/cloud/clouds/openstack.py
@@ -727,6 +727,8 @@ def create(vm_):
             )
             for private_ip in private:
                 private_ip = preferred_ip(vm_, [private_ip])
+                if private_ip is False:
+                    continue
                 if salt.utils.cloud.is_public_ip(private_ip):
                     log.warning('{0} is a public IP'.format(private_ip))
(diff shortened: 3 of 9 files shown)


isbm:
@mjura thanks for the PR! Could you please also provide a unit test for it?

mjura:
@isbm: I really would like to write this unit test for it, but unfortunately I have plenty other task to do. This change was cherry-picked from salt master branch. It was submitted and merged also without unit tests. Please look on this https://github.com/saltstack/salt/pull/39464. I've created also bug report about this https://bugzilla.suse.com/show_bug.cgi?id=1032749

isbm:
@mjura well, we all have plenty other tasks to do. Someone should finish your job then. The unit test isn't any hard to write, except the method is "private". But you still can access it.

isbm:
@dincamihai @pkazmierczak @meaksh do we need a test for this one? I'd say yes, because it is a change in IP filtering. Opinions?

dincamihai:
@isbm maybe we don't need to `Change "private" method to "hidden"` if we test the `create` wrapping function

@mjura how can we know that your cherry pick does the right thing and how can we know that this functionality will not be broken in the next update?

isbm:
@dincamihai in this case, yes, absolutely. I just thought we can make the UT more narrow.
@mjura I am +1 on @dincamihai 's question.

pkazmierczak:
I agree with @dincamihai on keeping "private", and I do think @isbm is right—we need a test.

isbm:
> Keep in mind that these changes will be included soon or later.

That is exactly why we need the test case for it! [This example](https://github.com/saltstack/salt/pull/40249#issuecomment-288737979) speaks all by itself.

mjura:
> I agree that a unit test for this would be helpful, but at the same time I would not expend too much time in case it turns hard to implement. Keep in mind that these changes will be included soon or later.

You took it out from my mouth. Thanks for saying this :)

mjura:
Probably Containers Team will spend more time and effort on developing salt-cloud and in upstream too. What I did last week was part of PoC during my Hack Week project about using salt-cloud with different Public Clouds providers. It looks that salt-cloud still have many different issues which can be addressed in upstream.

mjura:
It was also good exercise to learn about our development workflow for salt packages.

isbm:
@mjura but for us enterprise support isn't a hackweek, and what is here goes to SLE/ASM, and then runs somewhere in Sony, Toshiba or nuclear power plant. You know...

isbm:
> You took it out from my mouth. Thanks for saying this :)

He didn't meant that. :stuck_out_tongue:

isbm:
@mjura bump. Any tests coming anytime soon?

meaksh:
JFYI the target branch of this PR `openSUSE-2016.11.3` has been force pushed to get rid of 3929f63 which is now present on the source branch of this PR.

It's needed to rebase the source branch to get rid of that commit before this PR can be merged.

isbm:
We have to just close this. I am preparing the unit tests for this and they will go through the path via upstream, and so they will be picked up as usual.

isbm:
I am closing this. Tests are [pending in upstream](https://github.com/saltstack/salt/pull/40817) and we will pick that all up from there.

PR #26, opened 2017-04-10
2015.8.12: Set SHELL env var for salt-api.service

mateiw:
ssh `ProxyCommand` needs a valid `$SHELL` in order to execute the command.


isbm:
@mateiw please use `pkg/suse/salt-api.service` instead.

PR #27, opened 2017-04-10
2016.11.1: Set SHELL env var for salt-api.service

mateiw:
ssh `ProxyCommand` needs a valid `$SHELL` in order to execute the command.


isbm:
@mateiw  2016.11.1 is unsupported. Also this should go upstream for 2016.x And we do have `pkg/suse/salt-api.service`, why not use that one instead?

Closing as per https://github.com/openSUSE/salt/wiki/Workflow#submitting-patches

PR #28, opened 2017-04-10
2015.8.12: SUSE specific salt-service.api

mateiw:
Use a SUSE specific salt-service api instead of the generic `pkg/salt-service.api`
Includes changes from the package patches.
Adds SHELL env var for the salt-api.service. This is because `ssh -o ProxyCommand` needs a valid $SHELL in order to execute the given command.

### DIFF
--- a/pkg/suse/salt-api.service
+++ b/pkg/suse/salt-api.service
@@ -1,1 +1,15 @@
-../salt-api.service
+[Unit]
+Description=The Salt API
+After=network.target
+
+[Service]
+User=salt
+Type=simple
+Environment=SHELL=/bin/bash
+LimitNOFILE=8192
+ExecStart=/usr/bin/salt-api
+TimeoutStopSec=3
+
+[Install]
+WantedBy=multi-user.target
+


dincamihai:
[pkg/salt-api.service](https://github.com/openSUSE/salt/blob/openSUSE-2015.8.0/pkg/salt-api.service) has `After=syslog.target network.target`. Do we need the syslog.target here as well?

isbm:
@dincamihai I think so. In case someone wants to use syslog and watch all the failed messages there. :-)

dincamihai:
@mateiw I don't know how this works but is this new service file going to be automatically preferred over the pkg/salt-api.service or do we need to change some configuration?

mateiw:
@dincamihai added `After=syslog.target`.
No, it won't be used automatically. The spec file needs to be changed to install `pkg/suse/salt-api.service` instead of the generic one.

dincamihai:
@meaksh I've pointed to the pkg/salt-api.service file that exists on this branch because i believe that's what we used before introducing this new service file. To me it makes sense to keep everything from there since this PR only adds the SHELL env variable.
If there's a good reason to remove the `syslog.target`, we can do that as well. I would do it in a separate commit since these are global changes for the service.

meaksh:
@dincamihai - For me, your link is currently pointing to `pkg/salt-api.service` from `openSUSE-2015.8.0` branch instead of `openSUSE-2015.8.12` where this file is actually modified to remove this `syslog.target` prefix. 
Of course, I agree with you, we should keep everything as it on the current branch.

mateiw:
Ok, undone the last change.

dincamihai:
@mateiw thanks and sorry for the confusion

mateiw:
No problem @dincamihai

mcalmer:
@mateiw I think we need this also in 2016.11.3 ?

mateiw:
@mcalmer yes. I opened a PR upstream https://github.com/saltstack/salt/pull/40620. After it's merged it will need to be added also to our version.

PR #30, opened 2017-05-22
Fixed parsing of master minion returns when batching is enabled

grahamhayes:
This fixes https://bugzilla.suse.com/show_bug.cgi?id=1038814 for the CaaSP product, as we use the orchestration runner to install and update our install.

### DIFF
--- a/salt/states/saltmod.py
+++ b/salt/states/saltmod.py
@@ -300,7 +300,7 @@ def state(
             except KeyError:
                 m_state = False
             if m_state:
-                m_state = salt.utils.check_state_result(m_ret)
+                m_state = salt.utils.check_state_result(m_ret, recurse=True)
 
         if not m_state:
             if minion not in fail_minions:
@@ -309,9 +309,10 @@ def state(
             continue
         try:
             for state_item in six.itervalues(m_ret):
-                if 'changes' in state_item and state_item['changes']:
-                    changes[minion] = m_ret
-                    break
+                if isinstance(state_item, dict):
+                    if 'changes' in state_item and state_item['changes']:
+                        changes[minion] = m_ret
+                        break
             else:
                 no_change.add(minion)
         except AttributeError:


isbm:
Please open this PR in upstream instead! Thanks.

grahamhayes:
This is for the openSUSE product - for a team in SUSE using this package. an email was sent to the list about it.

This has been fixed upstream already - this just brings the patch back to our package.

ajaeger:
This is a backport of https://github.com/saltstack/salt/pull/40905, please reconsider, @isbm

isbm:
@ajaeger ah, OK. Although it is P1, yet since this is not an L3, I would ask where are the unit tests for it? Because we would like to have _every_ patch tested.

grahamhayes:
@isbm this is a clean backport of an upstream PR (I literally just `git cherry-pick <sha>`'d it on to this branch). I can add unit tests, but I would have to spend a lot of time figuring out how the salt test harness works, and they should really go upstream before we add them here.

isbm:
@grahamhayes That is the problem: "clean backport from upstream" != "clean support of Salt in SLE". You basically say to us to do these tests ourselves instead of you.

kkaempf:
Adding @moio since the title mentions "minion returns" and "batching" :wink:

isbm:
@kkaempf right. Although I know why, because I wrote that recursive part in Utils they're just using now... :smile:

pkazmierczak:
@isbm I vote UT too.

eotchi:
We **really** need this in the next maintenance update. Without it, the scalabity of our deployments is highly compromized. As the maintenance update is getting finalized as we speak, could we agree on having this accepted now and we commit to provide tests as soon as possible?

kiall:
Also, FYI, I spoke with Graham and he is currently looking at how the testing is done etc to write up a test case.

isbm:
@eotchi sure. Here only Unit test is needed, not an integration test. Salt Toaster will run them (UTs) as well. This can be done in an hour for a beginner. :-)

isbm:
@pkazmierczak @meaksh @dincamihai OK, @eotchi called this _really urgent_, so I am accepting this, as this is critical state. However, @grahamhayes owes us an UT for it. :wink:

PR #33, opened 2017-08-23
Notify systemd synchronously via NOTIFY_SOCKET (bsc#1053376)

renner:
### What does this PR do?

This patch is an attempt to make the notification of systemd more reliable in the fallback case of `systemd.daemon` python library not being available. It notifies the systemd `NOTIFY_SOCKET` synchronously instead of forking the `systemd-notify` command.

Forking the `systemd-notify` command is [known to be unreliable](https://bugzilla.redhat.com/show_bug.cgi?id=820448) at least with older versions of the kernel and/or systemd. When systemd receives the notification, the `systemd-notify` process may have already exited causing an error in the logs while waiting for a (90 seconds) timeout. This output is from a SLES12 system (systemd 210 and kernel 3.12.60-52.54-default):

```
Aug 15 11:34:50 client4-01 systemd[1]: Cannot find unit for notify message of PID 17264.
...
Aug 15 11:36:20 client4-01 systemd[1]: salt-minion.service start operation timed out. Terminating.
Aug 15 11:36:20 client4-01 salt-minion[17260]: [WARNING ] Minion received a SIGTERM. Exiting.
Aug 15 11:36:20 client4-01 salt-minion[17260]: The Salt Minion is shutdown. Minion received a SIGTERM. Exited.
Aug 15 11:36:20 client4-01 systemd[1]: Failed to start The Salt Minion.
Aug 15 11:36:20 client4-01 systemd[1]: Unit salt-minion.service entered failed state.
Aug 15 11:36:35 client4-01 systemd[1]: salt-minion.service holdoff time over, scheduling restart.
```

The patch is supposed to fix [bsc#1053376](https://bugzilla.suse.com/1053376) which is actually the same as [brc#820448](https://bugzilla.redhat.com/show_bug.cgi?id=820448). It was merged into upstream salt as https://github.com/saltstack/salt/pull/42986 and supposed to be forward-ported. The two commits in the upstream pull request were squashed together for this patch.

### Previous Behavior

In case the `systemd.daemon` library cannot be imported, systemd was notified asynchronously by forking the `systemd-notify` command.

### New Behavior

Notification of systemd is done **synchronously** via the `NOTIFY_SOCKET` and the python `socket` low level interface.

### Tests written?

No

### DIFF
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -15,6 +15,7 @@ import contextlib
 import subprocess
 import multiprocessing
 import multiprocessing.util
+import socket
 
 
 # Import salt libs
@@ -55,7 +56,20 @@ def notify_systemd():
         import systemd.daemon
     except ImportError:
         if salt.utils.which('systemd-notify') and systemd_notify_call('--booted'):
-            return systemd_notify_call('--ready')
+            # Notify systemd synchronously
+            notify_socket = os.getenv('NOTIFY_SOCKET')
+            if notify_socket:
+                # Handle abstract namespace socket
+                if notify_socket.startswith('@'):
+                    notify_socket = '\0{0}'.format(notify_socket[1:])
+                try:
+                    sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+                    sock.connect(notify_socket)
+                    sock.sendall('READY=1'.encode())
+                    sock.close()
+                except socket.error:
+                    return systemd_notify_call('--ready')
+                return True
         return False
 
     if systemd.daemon.booted():


renner:
@dincamihai and @brejoc: please have a look if this is as expected (I cannot really assign a reviewer it seems due to missing rights).

dincamihai:
@renner i know this was already accepted upstream but I just have one question.
Your code changes the behavior a bit and I just want to make sure it was intended.

Before `if salt.utils.which('systemd-notify') and systemd_notify_call('--booted')` was `True`, this was returned `systemd_notify_call('--ready')`
Now, `systemd_notify_call('--ready')` is returned only `if salt.utils.which('systemd-notify') and systemd_notify_call('--booted')` and `os.getenv('NOTIFY_SOCKET') is defined` and `socket.error is raised`. Was this intended?

renner:
@dincamihai: yes, this is exactly the intended behavior. Again:

1. if `systemd.daemon` library is there -> use it
2. if not: try to send the notification directly to the socket (using a call to `systemd-notify --booted` to check if this system is based on systemd)
3. if there is a `socket.error` while trying 2: fall back to calling `systemd-notify --booted` (which is known to be unreliable on older systemd/kernel versions but eventually will start the service after a 90 sec. timeout)

One assumption that I make: if we are on systemd but the socket is not defined (`os.getenv('NOTIFY_SOCKET')`), probably nothing will work as all the methods try to notify that socket. Therefore return `False`.

dincamihai:
@renner 
> if there is a socket.error while trying 2: fall back to calling `systemd-notify --booted`

I assume you mean `systemd_notify_call('--ready')`

Other than that it makes sense.
Feel free to merge it. Thanks!

isbm:
@renner I remember this, thanks for the PR! But I am not sure, did you committed it first to the upstream to the corresponding version? Normally we're here having no PRs at all, this repo is a tracker only as per [workflow](https://github.com/openSUSE/salt/wiki/Workflow#submitting-patches) and so we are only cherry-picking from the upstream the commits that are referenced in a corresponding bug in our Bugzilla.

If you did this all, then sorry asking, and I guess this PR is still not needed and can be closed, actually.

brejoc:
@isbm @dincamihai Yes, this has already been merged by upstream: https://github.com/saltstack/salt/pull/42986/

But actually @renner is following the [workflow](https://github.com/openSUSE/salt/wiki/Workflow#patch-validation) here. Shouldn't we create a PR first?

dincamihai:
@isbm it says here that we need to create PRs: https://github.com/openSUSE/salt/wiki/Workflow#patch-validation on point 2.

and especially in this case, @renner doesn't have rights to create branches or merge branches in opensuse/salt I would assume he can't commit directly either.

isbm:
@brejoc @renner ah, OK. Squash&merge then. :wink:

dincamihai:
@brejoc let's create the patch for OBS together.

isbm:
@brejoc @dincamihai :metal: :metal: :laughing:

renner:
Thanks guys! I am even more puzzled about the workflow now, but I already squashed the commits before making this PR. So it should be just a `git format-patch -1`, then adding the patch to the package and `osc sr` with the right target.

PR #34, opened 2017-08-24
yumpkg.list_pkg  and zypper.list_pkgs: add parameter for returned attribute selection (bsc#1052264)

brejoc:
### zypper

It adds a new optional parameter to list_pkg in the zypper module to return more data than the version (original reason is that for SUSE Manager integration we also need arch and install_date). Format is the same of existing method info_installed.

### yum

PR is merged upstream: https://github.com/saltstack/salt/pull/42310
PR is merged upstream: saltstack/salt#42718

### New Behavior

### Tests written?

Yes - existing tests have been adapted and a new one has been added.

### DIFF
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -5,6 +5,7 @@ Resources needed by pkg providers
 
 # Import python libs
 from __future__ import absolute_import
+import copy
 import fnmatch
 import logging
 import os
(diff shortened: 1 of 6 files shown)


brejoc:
I'm closing this in favor of #36, because in products:next we've got version `2017.7.1` and not `2017.7`. This caused some issues. @dincamihai was so kind to also create a branch for `2017.7.1`.

PR #35, opened 2017-08-24
yumpkg.list_pkg: add attr option to pkg.install (bsc#1052264)

brejoc:
### What does this PR do?

Merged upstream: https://github.com/saltstack/salt/pull/42718

### New Behavior

### Tests written?

Yes/No

Yes - existing tests have been adapted as well.

### DIFF
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -5,6 +5,7 @@ Resources needed by pkg providers
 
 # Import python libs
 from __future__ import absolute_import
+import copy
 import fnmatch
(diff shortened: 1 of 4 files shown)


brejoc:
@dincamihai Do we even have to created patches in this case? If I'm not mistaken, this would go to products:next. So if we are going to roll to `2017.7.2` for the next release there is a chance, that this might come from upstream anyways. Am I off track here?

brejoc:
Closing this because this has to be combined with the zypper changes in #34.

PR #36, opened 2017-08-25
yumpkg.list_pkg and zypper.list_pkgs: add parameter for returned attribute selection (bsc#1052264)

brejoc:
### zypper

It adds a new optional parameter to list_pkg in the zypper module to return more data than the version (original reason is that for SUSE Manager integration we also need arch and install_date). Format is the same of existing method info_installed.

### yum

PR is merged upstream: https://github.com/saltstack/salt/pull/42310
PR is merged upstream: saltstack/salt#42718

### New Behavior

### Tests written?

Yes - existing tests have been adapted and a new one has been added.

### DIFF
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -5,6 +5,7 @@ Resources needed by pkg providers
 
 # Import python libs
 from __future__ import absolute_import
+import copy
 import fnmatch
 import logging
 import os
@@ -306,3 +307,31 @@ def check_extra_requirements(pkgname, pkgver):
         return __salt__['pkg.check_extra_requirements'](pkgname, pkgver)
 
     return True
+
+
+def format_pkg_list(packages, versions_as_list, attr):
+    '''
+    Formats packages according to parameters for list_pkgs.
+    '''
+    ret = copy.deepcopy(packages)
+    if attr:
+        requested_attr = set(['version', 'arch', 'install_date', 'install_date_time_t'])
+
+        if attr != 'all':
+            requested_attr &= set(attr + ['version'])
+
+        for name in ret:
+            versions = []
+            for all_attr in ret[name]:
+                filtered_attr = {}
+                for key in requested_attr:
+                    filtered_attr[key] = all_attr[key]
+                versions.append(filtered_attr)
+            ret[name] = versions
+        return ret
+
+    for name in ret:
+        ret[name] = [d['version'] for d in ret[name]]
+    if not versions_as_list:
+        stringify(ret)
+    return ret
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -17,7 +17,6 @@ Support for YUM/DNF
 # Import python libs
 from __future__ import absolute_import
 import contextlib
-import copy
 import datetime
 import fnmatch
 import itertools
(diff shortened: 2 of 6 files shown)


brejoc:
@dincamihai SR is already submitted.

isbm:
@brejoc you mean that one that breaks tests? :wink:

brejoc:
@isbm That might very well be. Perhaps @moio should also take a look there.

brejoc:
@isbm But if I see this correctly, this is either the curl-error (https://github.com/SUSE/spacewalk/issues/1991) or has nothing to do with the changes made by @moio. Am I wrong?

`org.apache.http.conn.HttpHostConnectException: Connect to logstash.mgr.suse.de:9200 [logstash.mgr.suse.de/10.162.210.83] failed: Connection refused (Connection refused)`

`/salt/src/salt-2017.7.1/tests/integration/modules/test_timezone.py:28: AssertionError: 'ERROR: timedatectl failed: Failed to create bus connection: No such file or directory' not found in ['UTC', 'localtime']
/salt/src/salt-2017.7.1/tests/integration/runners/test_fileserver.py:174: AssertionError: False is not true`

`________________ ERROR collecting unit/modules/test_boto_vpc.py ________________
/salt/src/salt-2017.7.1/tests/unit/modules/test_boto_vpc.py:133: in <module>
    @skipIf(_has_required_moto() is False, 'The moto version must be >= to version {}. Installed: {}'.format(required_moto_version, _get_moto_version()))
/salt/src/salt-2017.7.1/tests/unit/modules/test_boto_vpc.py:106: in _get_moto_version
    return LooseVersion(moto.__version__)
E   NameError: global name 'moto' is not defined`

moio:
I lack context here, but I think my patch is only responsible for this error:

https://ci.suse.de/user/manager/my-views/view/Salt/view/default/job/manager-products-testing-suse-integration-rhel7/441/console

Specifically:
```
E     File "/usr/lib/python2.7/site-packages/salt/utils/pkg/rpm.py", line 105, in parse_pkginfo
E       install_date = datetime.datetime.utcfromtimestamp(int(install_time)).isoformat() + "Z"
E   ValueError: invalid literal for int() with base 10: '(none)'
```

IIUC this is because RPM on Red Hats can return `'(none)'` in absence of a value, and my patch did not check for it.

Here a correct example: https://github.com/moio/salt/blob/a49c41166032e367fa95dc2c427caef76f8ee868/salt/utils/pkg/rpm.py#L104
Here my failing code: https://github.com/moio/salt/blob/a49c41166032e367fa95dc2c427caef76f8ee868/salt/utils/pkg/rpm.py#L107

Unfortunately I will not be able to address this before Wednesday - so if you guys have time for it great, otherwise I will take care as soon as possible.

Thanks!

brejoc:
@moio Looks like we'd have to hurry up, if this should be in the next release. So what should be the value of `install_date`, when `install_time` is `None`?

Is there an other issue that needs to be adressed with this patch? You mentioned something like that somewhere else. Because we'd have to address this now.

moio:
`install_date` should be `None` if `install_time` is `'(none)'`.

The other issue is: https://github.com/saltstack/salt/pull/43195

brejoc:
Closing this in favor of #39.

PR #37, opened 2017-08-28
Revert "We don't have python-systemd, so notify can't work"

renner:
### What does this PR do?

This reverts commit 7917ec3a7efa9abda5932a0241070e0b048aa90d as notifying systemd should work now after the merge of #33. We can therefore simply remove the respective patch from our package:

```
tserong-suse.com-we-don-t-have-python-systemd-so-not.patch
```

### Previous Behavior

Service type for the `salt-master` systemd service was `simple`.

### New Behavior

Service type for the `salt-master` systemd service is `notify` as in upstream.

### Tests written?

No

### DIFF
--- a/pkg/salt-master.service
+++ b/pkg/salt-master.service
@@ -4,7 +4,8 @@ After=network.target
 
 [Service]
 LimitNOFILE=16384
-Type=simple
+Type=notify
+NotifyAccess=all
 ExecStart=/usr/bin/salt-master
 TasksMax=infinity
 


renner:
@dincamihai, @brejoc: Could you please have a look at this and let me know how to proceed with this one? Thanks!

mcalmer:
@renner just to be sure: This is a salt internal thing if we need this patch or not, or does it also depends on the systemd version we use on the system?

renner:
@mcalmer: The current Salt code (after #33) should work fine with any system based on sytemd, that's why this patch should not be needed anymore. It was furthermore changing only the master systemd unit file to type `simple`, but not the `salt-api` and `salt-minion` ones!

dincamihai:
@renner once we merge this we would need an SR to products:testing to remove the unneeded patch `tserong-suse.com-we-don-t-have-python-systemd-so-not.patch`
The tests would then run and once they are green I would just copy it to `products` and from there, with the OBS links in place there is nothing else to do.

renner:
Fine @dincamihai, I can make the SR once this is merged.

PR #56, opened 2017-11-24
Statefailure (L3 bsc#1068446)

isbm:
### What does this PR do?

Fixes broken logic for pillars. Problem is that Salt originally looking for error in `__pillar__` **and** any override of it. But if found, it pulls out only from `__pillar__`. Since nobody was using overrides, it went unnoticed till the moment when override pillar contains errors. Surely it crashes with `error = __pillar__['_errors']` key "`_errors`" was not found...

### Tests written?

Yes

@meaksh Here is the fix for our L3. I never pushed that upstream, because 2016.x ...

:bomb: **CAUTION:** before merge, it shall be squashed into one commit! :fire: :fire: :fire:

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -103,12 +103,7 @@ def _check_pillar(kwargs, pillar=None):
     Check the pillar for errors, refuse to run the state if there are errors
     in the pillar and return the pillar errors
     '''
-    if kwargs.get('force'):
-        return True
-    pillar_dict = pillar if pillar is not None else __pillar__
-    if '_errors' in pillar_dict:
-        return False
-    return True
+    return None if kwargs.get('force') else (pillar or {}).get('_errors', __pillar__.get('_errors')) or None
 
 
 def _wait(jid):
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -380,10 +380,10 @@ def template(tem, queue=False, **kwargs):
         return conflict
     st_ = salt.state.HighState(__opts__, context=__context__)
 
-    if not _check_pillar(kwargs, st_.opts['pillar']):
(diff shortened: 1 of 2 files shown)


meaksh:
FYI - according to Salt lifecycle [1] the 2016.11 branch is still accepting bug fixes until November 30. You're still on time to push it to upstream 2016.11.

[1] https://saltstack.com/product-support-lifecycle/

salt/modules/state.py
```diff
@@ -98,17 +98,16 @@ def _set_retcode(ret):
...
+    :param pillar: external pillar
+    :return: None or an error message
     '''
-    if kwargs.get('force'):
-        return True
-    pillar_dict = pillar if pillar is not None else __pillar__
-    if '_errors' in pillar_dict:
-        return False
-    return True
+    return None if kwargs.get('force') else (pillar or {}).get('_errors', __pillar__.get('_errors')) or None
```
meaksh: I would probably unpack this one-line into:

```python
if kwargs.get('force'):
    return None
_errors = (pillar or {}).get('_errors', __pillar__.get('_errors')
return _errors if _errors else None
```

That would make more intuitive to get the actual behavior of this function. I'm not a big fan of compressing too much logic into one line as it might lead into confusion.

> Sparse is better than dense. (https://www.python.org/dev/peps/pep-0020/)

isbm: I don't like several returns. Besides this is just a port from what is in 2017+

isbm:
@meaksh I would not really care about sending PRs anymore to upstream of this version, although one week left. They are unable to even merge `develop` and 2017 properly. So let they focus on these instead.

isbm:
Merged.

PR #57, opened 2017-11-24
Cherry-pick changes to run salt-master as user salt.

clanig:
### What does this PR do?
Run salt-master as user `salt`

https://github.com/SUSE/spacewalk/issues/2422

### Previous Behavior
salt-master ran as `root`.

### New Behavior
salt-master should run as `salt`.

### Tests written?
No

### DIFF
--- a/pkg/salt-api.service
+++ b/pkg/salt-api.service
@@ -6,6 +6,7 @@ After=network.target
 [Service]
 Type=notify
 NotifyAccess=all
+User=salt
 LimitNOFILE=8192
 ExecStart=/usr/bin/salt-api
 TimeoutStopSec=3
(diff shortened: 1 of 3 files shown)


pkg/salt-api.service
```diff
@@ -4,8 +4,8 @@ Documentation=man:salt-api(1) file:///usr/share/doc/salt/html/contents.html http
 After=network.target
 
 [Service]
-Type=notify
```
meaksh: I just realized that we should split this commit. On the past we removed the `Type=notify` and put `Type=simple' but we removed that few months ago, so according to this, this commit should only set `User=salt` here.

PR #62, opened 2017-12-28
Feat: add grain for all FQDNs (bsc#1063419)

mbologna:
This PR adds a grain named fqdns to the grains.
fqdns represents all the FQDNs known for the system on all available interfaces (excluding lo).

Note: hostname != FQDN

hostname is the UNIX name of the machine. A machine can have one and only one hostname.
FQDN is host.domain that resolves to an IP address that the machines is answering to.
A machine can have 1+ FQDNs.

Upstream PR:
https://github.com/saltstack/salt/pull/45060

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### DIFF
--- a/pkg/suse/salt-common.logrotate
+++ b/pkg/suse/salt-common.logrotate
@@ -8,6 +8,7 @@
 }
 
 /var/log/salt/minion {
+	su root root
 	weekly
 	missingok
 	rotate 7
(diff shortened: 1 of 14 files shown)


mbologna:
For future reference: https://github.com/SUSE/spacewalk/issues/2567#issuecomment-354142148

mbologna:
Re-opened as https://github.com/openSUSE/salt/pull/63

PR #73, opened 2018-04-11
[openSUSE-2018.3.0] backport #46890 - fall back to PyMySQL

MaximilianMeister:
### What does this PR do?
adds a fallback to a drop in library. the same way is already done in modules

NOTE: I guess it's better to wait until https://github.com/saltstack/salt/pull/46890 is merged, in case there are more changes necessary

### DIFF
--- a/salt/auth/mysql.py
+++ b/salt/auth/mysql.py
@@ -55,10 +55,29 @@ import logging
 log = logging.getLogger(__name__)
 
 try:
+    # Trying to import MySQLdb
     import MySQLdb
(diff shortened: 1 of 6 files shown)


MaximilianMeister:
also note that this is also not yet backported to `2018.3.0` upstream, only on `develop` but the patch applied fine

MaximilianMeister:
@dincamihai upstream got merged. please have a look. shall I backport it upstream first to `2018.3.0` ?

PR #76, opened 2018-04-12
Fix for EC2 Rate Limit Failures

edlane:
### What does this PR do?
fix for ec2 rate limit failures described here:
https://bugzilla.suse.com/show_bug.cgi?id=1088888

See:
https://bugzilla.suse.com/show_bug.cgi?id=1088888

### Previous Behavior
See:
https://bugzilla.suse.com/show_bug.cgi?id=1088888

### New Behavior
See:
https://bugzilla.suse.com/show_bug.cgi?id=1088888

### Tests written?
Yes
See:
https://bugzilla.suse.com/show_bug.cgi?id=1088888

### DIFF
--- a/salt/utils/aws.py
+++ b/salt/utils/aws.py
@@ -20,6 +20,7 @@ import hmac
 import logging
 import salt.config
 import re
+import random
 
 # Import Salt libs
 import salt.utils.xmlutil as xml
@@ -449,10 +450,11 @@ def query(params=None, setname=None, requesturl=None, location=None,
         )
         headers = {}
 
-    attempts = 5
-    while attempts > 0:
-        LOG.debug('AWS Request: {0}'.format(requesturl))
-        LOG.trace('AWS Request Parameters: {0}'.format(params_with_headers))
+    MAX_RETRIES = 6
+    attempts = 0
+    while attempts < MAX_RETRIES:
+        log.debug('AWS Request: %s', requesturl)
+        log.trace('AWS Request Parameters: %s', params_with_headers)
         try:
             result = requests.get(requesturl, headers=headers, params=params_with_headers)
             LOG.debug(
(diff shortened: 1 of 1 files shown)


salt/utils/aws.py
```diff
@@ -449,10 +450,11 @@ def query(params=None, setname=None, requesturl=None, location=None,
...
 
-    attempts = 5
-    while attempts > 0:
-        LOG.debug('AWS Request: {0}'.format(requesturl))
-        LOG.trace('AWS Request Parameters: {0}'.format(params_with_headers))
+    MAX_RETRIES = 6
+    attempts = 0
+    while attempts < MAX_RETRIES:
+        log.debug('AWS Request: %s', requesturl)
+        log.trace('AWS Request Parameters: %s', params_with_headers)
```
brejoc: Perhaps I'm overlooking something, but where is that lowercase `log` coming from?

salt/utils/aws.py
```diff
@@ -475,16 +477,24 @@ def query(params=None, setname=None, requesturl=None, location=None,
 
             # check to see if we should retry the query
             err_code = data.get('Errors', {}).get('Error', {}).get('Code', '')
-            if attempts > 0 and err_code and err_code in AWS_RETRY_CODES:
-                attempts -= 1
+            if attempts < MAX_RETRIES and err_code and err_code in AWS_RETRY_CODES:
```
brejoc: Couldn't we can skip the check for `attempts < MAX_RETRIES` here, since this is the condition for the while loop?

brejoc:
@edlane Looks like this is the upstream PR, right? 👉 https://github.com/saltstack/salt/pull/47211

jgleissner:
Any update on this?

edlane:
@brejoc  Yes, that is the upstream (SaltStack) PR.
I was asked by Robert to also submit a PR to upsteam **openSUSE** 2018.3
I was waiting for your approval/merge here before proceeding on to other repos.

I will go back and include your suggested changes to the SaltStack (develop) PR with your approval/merge here and I will include you as the source -- you should get the credit for these changes.

brejoc:
@edlane Usually - if it is possible and upstream wants the changes - we push the changes upstream first and then just cherry-pick (and sometimes backport) from there this way we always align with what upstream is doing and won't have to maintain too many things ourselves. So pushing this to upstream first was spot on. Thanks! If there are still changes left, that haven't found its was upstream I would very much appreciate it if you'd push them upstream also. No need to mention me anywhere. I was just asking questions. 😃

Meanwhile I'll merge this and generate a patch for `products` later today. Thanks for the work!

brejoc:
@edlane Since this is done now, would you please open up the PR for 2018.3.0? Thanks!

PR #82, opened 2018-05-11
[openSUSE-2018.3] enable passing a unix_socket for mysql returners (bsc#1091371) 

MaximilianMeister:
quick fix for:
  https://bugzilla.suse.com/show_bug.cgi?id=1091371

the upstream patch will go through some bigger refactoring of
the mysql drivers to be cleaner

this patch should only be temporary and can be dropped again once
the refactor is done upstream


MaximilianMeister:
@isbm thanks, i plan to pick the task up again this week, and push it to a new branch, i'm just currently a bit under water with caasp tasks.

PR #72, opened 2018-04-06
Initial PR for backport of AzureARM from Salt 2018.3 to OpenSUSE-Salt 2016.11.4

edlane:
## NOTICE: This PR should be considered preliminary...
While it has been tested WRT AzureARM salt-cloud functionality it should be given more scrutiny and widespread Salt test coverage before acceptance.

### What does this PR do?
provides initial backport of Salt's AzureARM cloud provider from 2018.3 to 2016.11.4

### What issues does this PR fix or reference?
see https://bugzilla.suse.com/show_bug.cgi?id=1087342

### Previous Behavior
AzureARM was not supported in 2016.11.4

### New Behavior
AzureARM IS supported in 2016.11.4

### Tests written?
Yes -- see Testcase referenced in attachments here:
https://bugzilla.suse.com/show_bug.cgi?id=1087342

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

PR #88, opened 2018-06-05
salt-cloud: AzureARM 2018.3 backport fixes

jgleissner:
Modified salt/utils/cloud.py to work with previous versions of some utils modules,
as replacing them with 2018.3 versions causes breakage.
Removed salt/utils/crypt.py (added by previous backport commit)
Removed salt/utils/yaml.py (added by previous backport commit)
Reverted salt/utils/yamldumper.py and salt/utils/yamlloader.py to pre-backport
versions.

### What does this PR do?

This PR reverts some of the changes in salt/utils/ the previous [AzureARM backport](https://github.com/openSUSE/salt/pull/72) did.

I ran into issue with the previous backport (broken orchestration on CaaSP). The resulting modules in salt/utils/ are now closer to original 2016.11 and should be less likely to break things. I tested it successfully in a CaaSP context.


isbm:
@jgleissner what was the reason to _remove_ the entire `crypt` function from utils and how this would affect your part?

PR #101, opened 2018-07-19
Fix #46492: file.blockreplace throws IndexError.

bergmannf:
This was discovered to affect SUSE CaaS Platform 3, during a test-upgrade of salt
to version 2018.3.0. The update orchestration uses this module and breaks.

### What does this PR do?

It fixes the upstream issue saltstack/salt#46492 (also fixes [bsc#1101812](https://bugzilla.suse.com/show_bug.cgi?id=1101812)).

saltstack/salt#46492

### Previous Behavior

Does not verify that split_content actually has a value.

### New Behavior
Does verify that split_content actually has a value.

### Tests written?

No

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -2560,7 +2560,7 @@ def blockreplace(path,
 
                         # Trim any trailing new lines to avoid unwanted
                         # additional new lines
-                        while not split_content[-1]:
+                        while split_content and not split_content[-1]:
                             split_content.pop()
 
                         # push new block content in file


meaksh:
**IMPORTANT NOTE**

Contributions to Salt should be always made upstream. So we can also create a PR to this repository to bring those changes to our SUSE Salt packages if needed. (See contributions guideline for Salt on SUSE here: https://github.com/openSUSE/salt/wiki/Workflow#policies)

The upstream PR for this changes is missing, but since we're preparing maintenance update for Salt 2018.3.0 this week, I'm this PR now and will wait for @bergmannf to create the upstream PR also together with https://github.com/openSUSE/salt/pull/102

Thanks for the fix @bergmannf ! :+1:

bergmannf:
@meaksh Thanks for merging the PR!

I checked with the upstream version and it seems they have replaced the whole block in the current `2018.3` branch: https://github.com/bergmannf/salt-2/blob/2018.3/salt/modules/file.py#L2631

I hope once we rebase our salt-version to that one we should no longer need this fix.

PR #108, opened 2018-09-05
fix azurearm dependencies on SLES 12 SP3 (bsc#1107333)

jgleissner:
The cloud azurearm driver requires azure-cli, even though it is not used at all. It also requires the multi-api version of azure-storage, which is not available on SLES 12 SP3 but works fine with plain azure-storage. This commit removes the check for azure-cli and changes the check for azure-storage.

### Previous Behavior

Throws exception on SLES 12 SP3 when salt-cloud is invoked with azurearm backend configured.

### New Behavior
Passes dependency checks on SLES 12 SP3.

### Tests written?

No


jgleissner:
I think I've opened this against the wrong branch. Is openSUSE-2018.3.0 the right one?

isbm:
@jgleissner please reference here upstream PR.

PR #109, opened 2018-09-07
Support use of GCE instance credentials

jgleissner:
### What does this PR do?
Allow the user to configure an empty string as `service_account_private_key` (GCE specific cloud configuration parameter). The effect is that libcloud will use instance credentials for authentication, which is used by CaaSP in GCE.

### Previous Behavior
Running salt-cloud in GCE using instance credentials not possible.

### New Behavior
Running salt-cloud in GCE using instance credentials possible.

### Tests written?
No (but manually tested).

### DIFF
--- a/salt/cloud/clouds/azurearm.py
+++ b/salt/cloud/clouds/azurearm.py
@@ -25,6 +25,9 @@ The Azure cloud module is used to control access to Microsoft Azure
     * ``client_id``
     * ``secret``
 
+    if using MSI-style authentication:
+    * ``subscription_id``
+
 Example ``/etc/salt/cloud.providers`` or
 ``/etc/salt/cloud.providers.d/azure.conf`` configuration:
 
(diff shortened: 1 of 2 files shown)


isbm:
@jgleissner what is the PR reference in the upstream? Or there is none?

jgleissner:
@isbm I haven't opened a PR upstream. I just checked the upstream repo and the develop branch has the same limitation. Should I simply open the same RP upstream, or do you think this should be done differently?

isbm:
@jgleissner yes, this has to go first to the upstream and accepted there. Obviously, if there is a limitation everywhere and we are releasing the upstream package, your update will vanish in the next release.

jgleissner:
I've opened a PR upstream: https://github.com/saltstack/salt/pull/49605

meaksh:
Merged, thanks @jgleissner !

jgleissner:
Thanks @meaksh. The upstream PR is against the develop branch. Should I openen it against 2018.3 as well?

meaksh:
@jgleissner - Hmm, the upstream PR is actually a bugfix right? I think upstream would accept that fix into `2018.3` branch. :+1: 

That would be nice, so we don't need to keep this patch in our OBS until the next Salt major release.

jgleissner:
@meaksh I wouldn't say it's a bug fix per se, but the change is small and unintrusive. I'll open a PR against 2018.3 once they're fine with the patch.

PR #110, opened 2018-09-11
fix azurearm dependencies on SLES 12 SP3 (bsc#1107333)

jgleissner:
### What does this PR do?
The cloud azurearm driver requires azure-cli, even though it is not used at all. It also requires the multi-api version of azure-storage, which is not available on SLES 12 SP3 but works fine with plain azure-storage. This commit removes the check for azure-cli and changes the check for azure-storage.

### Previous Behavior
Throws exception on SLES 12 SP3 when salt-cloud is invoked with azurearm backend configured.

### New Behavior
Passes dependency checks on SLES 12 SP3.

### Tests written?
No

### DIFF
--- a/salt/states/linux_acl.py
+++ b/salt/states/linux_acl.py
@@ -66,7 +66,7 @@ def present(name, acl_type, acl_name='', perms='', recurse=False):
         ret['result'] = False
         return ret
 
-    __current_perms = __salt__['acl.getfacl'](name)
+    __current_perms = __salt__['acl.getfacl'](name, recursive=recurse)
 
     if acl_type.startswith(('d:', 'default:')):
         _acl_type = ':'.join(acl_type.split(':')[1:])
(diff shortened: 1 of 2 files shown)


isbm:
@jgleissner https://github.com/openSUSE/salt/pull/108#issuecomment-420249903

jgleissner:
@isbm Sorry, not sure how the MSI stuff slipped into this PR. I'll fix it.

Also, I didn't open an upstream PR because this patch basically just adjusts salt 2018.3 to azure packages that are available in SLES 12. SLES 15 and openSUSE have the multiapi azure-storage package and also the cli package and don't need this (although the dependency on azure-cli is superfluous and perhaps should be removed). This also may mean that this isn't the right place for this kind of change, in which case please advise what I need to do.

isbm:
@jgleissner We do not really differ between SLE12 and SLE15. We differ by Salt version instead. So the fix should figure out the environment it is running.

PR #112, opened 2018-09-21
fix azurearm deps bsc#1107333

jgleissner:
### What does this PR do?
This PR removes the dependecy of the azurearm driver on azure-cli (which is not used) and allows using the official azure-storage-sdk in case the multiapi package is not available (as it is the case on SLES 12 currently). The changes were merged upstream (https://github.com/saltstack/salt/pull/49696) and cherry-picked from there.


jgleissner:
This was actually merged with commit f8b83b9277348190276f5aaa769c05ec2e90b0c6. Closing.

PR #114, opened 2018-12-04
Fix usermod options for SLE11 (bsc#1117017)

isbm:
### What does this PR do?

Bugfix.

bsc#1117017

L3: salt-minion error while bootstrapping sles11sp4:
https://bugzilla.suse.com/show_bug.cgi?id=1117017

### Previous Behavior

usermod on SLE11 does not accept `-a`.


brejoc:
Waiting for feedback from the customer.

PR #116, opened 2018-12-13
Return the expected powerpc os arch (bsc#1117995)

dincamihai:
### What does this PR do?
Fixes https://bugzilla.suse.com/show_bug.cgi?id=1117995

`grains.get osarch` returns now powerpc64le which comes from `rpm --eval "%{_host_cpu}"`, before is was the same as `cpuarch` grain (ppc64le)

This bug was triggered by the following changes:
  - https://github.com/SUSE/salt-board/issues/119
  - https://github.com/saltstack/salt/commit/746d2be3902fd311b34cd32259cfdeac2508d800 https://bugzilla.suse.com/show_bug.cgi?id=1114197

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/utils/pkg/rpm.py
+++ b/salt/utils/pkg/rpm.py
@@ -54,8 +54,11 @@ def get_osarch():
             stderr=subprocess.PIPE).communicate()[0]
     else:
         ret = ''.join(list(filter(None, platform.uname()[-2:]))[-1:])
-
-    return salt.utils.stringutils.to_str(ret).strip() or 'unknown'
+    ret = salt.utils.stringutils.to_str(ret).strip() or 'unknown'
+    ARCH_FIXES_MAPPING = {
+        "powerpc64le": "ppc64le"
+    }
+    return ARCH_FIXES_MAPPING.get(ret, ret)
 
 
 def check_32(arch, osarch=None):


salt/utils/pkg/rpm.py
```diff
@@ -54,8 +54,11 @@ def get_osarch():
             stderr=subprocess.PIPE).communicate()[0]
     else:
         ret = ''.join(list(filter(None, platform.uname()[-2:]))[-1:])
-
-    return salt.utils.stringutils.to_str(ret).strip() or 'unknown'
+    ret = salt.utils.stringutils.to_str(ret).strip() or 'unknown'
+    ARCH_FIXES_MAPPING = {
+        "powerpc64le": "ppc64le"
+    }
+    return ARCH_FIXES_MAPPING.get(ret, ret)
```
isbm: The `powerpc64le` is actually a [mess on fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1461288) as well, seems like. Does `%{_arch}` macro returns also wrong?

isbm: @dincamihai Maybe that makes sense not to have such dict, but rather use `_arch` in grain instead of `_host_cpu`? I believe the `_host_cpu` was wrong idea from the beginning, but fires back only on PPC, since on x86 it is the same thing.

dincamihai: @isbm that could make sense, yes. But we have to be careful here because we are talking about changing core grains. If this _arch macro works on all platform (without changing the currently returned arch) than it would be a good idea to use it. But **only if we don't change the currently returned grain**

dincamihai: @isbm the dictionary is a fix for the PTF. As I already said, the long term solution could be different. (Maybe using _arch macro)

isbm: >  If this _arch macro works on all platform (without changing the currently returned arch) than it would be a good idea to use it.

That's right. I would be still voting for PTF the `_arch` instead. Thing is: on x86 this will work 100%, as it is the same thing:

```
$ rpm --eval '%{_arch}'
x86_64
$ rpm --eval '%{_host_cpu}'                                                                                                                            
x86_64
```

Second, the function is about to get _architecture_: `get_osarch`. So probably the very original code by terminalmage on 2016-02-16 wasn't right at the beginning. This is why I would be still try PTF `_arch` instead, because this should be tested anyway and this is a very good chance to do so.

Of course, I am not insisting on anything.

dincamihai: @isbm I see your point. It it sounds good to use _arch here in the get_osarch but I would argue that the PTF is meant to fix the problem for the customer as fast as possible since the customer might be in a critical situation (they can't onboard the minion). Let's test this _arch macro and when we are sure that that it doesn't change the grain, we can put it in the next release.

kkaempf: What does the `rpm` maintainer say about this ? Did anyone ask ?
```
power-novalink:~ #  rpm --eval '%{_arch}'
ppc64le
power-novalink:~ # rpm --eval '%{_host_cpu}'
powerpc64le
```

dincamihai: @kkaempf we did not ask. Do you know how we can contact the maintainer?

kkaempf: As with any other package:

- https://build.suse.de
- search for `rpm`
- find e.g. https://build.suse.de/package/show/SUSE:SLE-15:GA/rpm
- click on `Users`

dincamihai: @kkaempf thanks

@mlschroe can you please help us here? We would like to know what the _arch rpm macro would return on different platform. Thank you!

dincamihai:
Merging the workaround since we need the fix in the next release: https://github.com/juliogonzalez

meaksh:
@dincamihai can we safely remove this branch?

dincamihai:
@meaksh done

PR #123, opened 2019-02-08
Fix regression in dynamic pillarenv (bsc#1124277)

dincamihai:
Proposed patch for https://bugzilla.novell.com/show_bug.cgi?id=1124277

Upstream code is different (probably because of other backportings): https://github.com/saltstack/salt/blob/2018.3/salt/utils/gitfs.py#L2992-L3019

Just using the upstream implementation of that function doesn't work because it needs some changes to code outside the function. Since it is hard to establish what exactly needs to be ported from upstream, I have decided to produce a patch that fixes the problem we have even though it's not exactly as upstream.

I hope to unify with upstream at a later time.


dincamihai:
Thanks!

PR #124, opened 2019-02-12
[WIP] Fix Suse os_family grain in suse_apache module (bsc#1125101)

dincamihai:
### What does this PR do?
Fix https://bugzilla.novell.com/show_bug.cgi?id=1125101

### DIFF
--- a/salt/modules/suse_apache.py
+++ b/salt/modules/suse_apache.py
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
-    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'SUSE':
+    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'Suse':
         return __virtualname__
     return (False, 'apache execution module not loaded: apache not installed.')
 


aboe76:
@dincamihai nice find, thanks for fixing this

brejoc:
@aboe76 Sorry, but this is not how it works. This PR is still marked was work-in-progress [WIP] and shouldn't be merged. On top of that only members of the SUSE Salt team should merge in this repo. I don't even know why you had the permissions in the first place. :thinking:

We are only using this repo to build our packages. That's why write permissions should be restricted to the team. So no worries, but we'll have to revert this and check your permissions! :smile:

brejoc:
@aboe76 I've just noticed that you are a member of the salt team and you might have worked on openSUSE Salt in the past. So welcome back! :smile: 

I'm not sure how things locked back in the day, so I'd like to get you up-to-date. The openSUSE-xxx branches are used for the corresponding versions we are building in OBS. `openSUSE-2018.3.0` is the current stable release. `openSUSE-2019.2.0rc2` will be the next release. Everything that is being merged here should come from upstream, so that we don't have to maintain it ourselves. In order to get all of the patches to OBS, we use an other repo called [salt-packaging](https://github.com/openSUSE/salt-packaging). But to make this work every PR has to be squashed (into one commit) and we don't want the merge commits. Otherwise the [git formatting tool](https://github.com/openSUSE/salt/wiki/Using-git-format-pkg-patch) gets confused. I know this workflow is a bit special - so if you'd like to start contributing again, I'd suggest to get together to give you a quick demo. Because this is only half of the story. Things in OBS also got a little bit restructured. I'd be happy to hear form you!

PR #130, opened 2019-03-07
[WIP] Reload modules in state.apply bsc#1125444

dincamihai:
### What does this PR do?
Updated custom modules code is not used in the following state.apply.
Fix https://bugzilla.suse.com/show_bug.cgi?id=1125444

NOTE: we don't know if this is the correct way to fix the problem

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -1921,6 +1921,8 @@ class State(object):
 
         if 'provider' in low:
             self.load_modules()
+        else:
+            self.load_modules()
 
         if low.get('__prereq__'):
             low['__prereq__'] = False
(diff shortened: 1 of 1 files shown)


dincamihai:
> @dincamihai @meaksh can we get this PR merged downstream so we can have a pkg in caasp? tia
> 
> Imho for the best approach upstream we can backport this later on, but this is enough currently to fix our issues

The problem is we don't know if this is going to have a negative impact on SUSE Manager or not.

dincamihai:
Since there is nothing that requires this fix, I'm closing the PR.

PR #133, opened 2019-03-11
sync modules always. This workaround caasp issue

MalloZup:
### What does this PR do?
sync modules always. This workaround caasp issue


MalloZup:
@meaksh @dincamihai since in caasp we don't own a specific salt-pkg, but we use this one, we need the fix `downstream` asap and then we could have the upstream fix later on

MalloZup:
ok closing in favor of https://github.com/openSUSE/salt/pull/133

PR #137, opened 2019-03-29
[WIP] Set UUID from XEN sysfs if not set before

hustodemon:
### What does this PR do?

On XEN PV guest, there is no bios and commands like dmidecode or smbios
can't be used for determining UUID of the machine. XEN exports this
value in /sys/hypervisor/uuid file, when XEN_SYS_HYPERVISOR kernel
option is enabled and this patch makes use of it.

### Previous Behavior
On XEN PV guests, the uuid grain was not set.

### New Behavior
The uuid grain is set on XEN PNV guests

### Tests written?

Not **yet**

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -2451,6 +2451,17 @@ def _hw_data(osdata):
                 grains['product'] = res.group(1).strip().replace("'", "")
                 break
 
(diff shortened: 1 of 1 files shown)


SchoolGuy:
This PR is dead for a very long time. I will change the base against the new branch but will additionally close it.

PR #138, opened 2019-04-10
Open suse 2016.11.10 port optimization order

dincamihai:
### What does this PR do?
Fixes the missing `optimization_order` issue:

the test.opts_pkg returns the options from the sles11sp4(salt-2016.11.10 in the salt-ssh thin) but does not have 'optimization_order' parameter (the default [0, 1, 2]) - this causes the empty lazyloader.file_mapping issue which causes the 'state.apply' KeyError issue on the master https://github.com/openSUSE/salt/blob/openSUSE-2019.2.0/salt/client/ssh/__init__.py#L1159

Porting https://github.com/saltstack/salt/commit/6fc8da5bab37d4e250e01bc220d6153be2018076

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -264,6 +264,9 @@ VALID_OPTS = {
     # The type of hashing algorithm to use when doing file comparisons
     'hash_type': str,
 
+    # Order of preference for optimized .pyc files (PY3 only)
+    'optimization_order': list,
+
     # Refuse to load these modules
     'disable_modules': list,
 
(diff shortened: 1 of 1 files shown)


dincamihai:
Any objections to fixing it in this way?

brejoc:
I have no clue what's going on here. What does this do?

isbm:
@dincamihai yeah that would be correct way in this case.
@brejoc this has to do how LazyLoader is loading sources. E.g. on Py3 they want first (or even only) load only `.pyc` files from `__pycache__`.

dincamihai:
Merged here: https://github.com/openSUSE/salt/commit/48cf02e7af98a7dfe9919c5aeacf65f218663472
Closing PR.

PR #139, opened 2019-04-11
2018.3.0 fix async batch race conditions

dincamihai:
### What does this PR do?
Fixing race condition on async batching.
Fix batch termination.

### Previous Behavior
Adding minions to `active` tracking list and `batch scheduling` could be in a race condition.
Batch was not terminated at the right time.

### New Behavior
Add minions to the list before scheduling the batch.
Terminate batch if there is no minion `waiting` or `currently running`.

### Tests written?

Yes


dincamihai:
@chiaradiamarcelo this should fix the race conditions and the batch termination

PR #145, opened 2019-05-23
make sure not to send invalid information

brejoc:
### What does this PR do?

Upsteam PR can be found here: https://github.com/saltstack/salt/pull/47280

Localectl outputs n/a sometimes when stuff is not set, instead of just not
outputting anything.  Actually put None in the variable when parsing it, and do
not add it to the localectl set command.

update localemod tests

bsc#1128974

### Previous Behavior

locale module was not working

### New Behavior

locale module should be working again


brejoc:
No need to rebase this to 2019.2.0, since it's already in there and we no longer need it for 2018.3.0.

PR #146, opened 2019-05-29
[WIP] batch.py: avoid exception when minion does not respond (bsc#1135507)

dincamihai:
Porting: https://github.com/saltstack/salt/pull/53159
---

We have several issues reporting that salt is throwing exception when
the minion does not respond. This change avoid the exception adding a
default data to the minion when it fails to respond. This patch based
on the patch suggested by @roskens.

Issues #46876 #48509 #50238
bsc#1135507

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>

### DIFF
--- a/salt/cli/batch.py
+++ b/salt/cli/batch.py
@@ -315,6 +315,11 @@ class Batch(object):
                     if self.opts.get('failhard') and data['ret']['retcode'] > 0:
                         failhard = True
 
+                # avoid an exception if the minion does not respond.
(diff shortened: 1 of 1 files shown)


dincamihai:
@jvanz this is a port of your PR to the upstream `develop` branch. In order to bring your change to our package I need to port your PR here to version 2018.3.0 which is the salt version the bug was reported agains. I hope this clarifies it.

PR #148, opened 2019-06-07
Open suse yomi

aplanas:
### What does this PR do?

Provide the missing features required for Yomi

### Tests written?

Yes


aplanas:
Ok. Lets try this again.

This patch is the set of commits that are in develop (and in Neon) that are missing in out package.

I executed the toaster tests, and I detected that the base branch (2019.2.0) contains already broken tests in the unit tests and in the suse.tests.

I detected the reason for one unit tests, related with debian, and I added the fix in this PR. But there is still one bug in zypperpkg (not related with this PR), that make 'tests/unit/modules/test_zypperpkg.py::ZypperTestCase::test_list_pkgs_with_attr' fail. The problem is in the code, not in the tests as is the same test that is in develop, but in develop the test is green.

For suse.tests, before and after the code in this PR we have this:

6 failed, 63 passed, 37 skipped, 21 xfailed, 3 xpassed

aplanas:
Saltstack integration tests do not show any regression. Is ready for review.

meaksh:
Closing in favor of https://github.com/openSUSE/salt/pull/161

PR #163, opened 2019-07-24
cmdmod: fix runas and group in run_chroot

aplanas:
The parameters runas and group for cmdmod.run() will change the efective
user and group before executing the command. But in a chroot environment is
expected that the change happends inside the chroot, not outside, as the
user and groups are refering to objects that can only exist inside the
environment.

This patch add the userspec parameter to the chroot command, to change
the user in the correct place.

(cherry picked from commit f0434aaeeee3ace4e3fc65c04e69984f08b2541e)

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -3064,13 +3064,19 @@ def run_chroot(root,
 
     if isinstance(cmd, (list, tuple)):
         cmd = ' '.join([six.text_type(i) for i in cmd])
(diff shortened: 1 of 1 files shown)


meaksh:
Hey @aplanas, let's wait until we have some feedback from upstream here. I'll keep the eye on this :+1:

meaksh:
https://github.com/saltstack/salt/pull/53992

aplanas:
Sure, to have a consistent workflow I make sure to submit to the three places: develop, neon and openSUSE at the same time, so I do not forget any missing piece

PR #165, opened 2019-07-24
Accumulated changes required for Yomi

aplanas:
### What does this PR do?

Partial accumulated changes required for Yomi:

* cmdmod: fix runas and group in run_chroot (https://github.com/saltstack/salt/pull/53992)
* Fix kwargs iteration for chroot.call (https://github.com/saltstack/salt/pull/53996)
* Some improvements and fixes to the freezer module (https://github.com/saltstack/salt/pull/54022)
* zypperpkg: fix pkg.list_pkgs cache (https://github.com/saltstack/salt/pull/54024)

### Tests written?

Yes

### DIFF
--- a/salt/modules/chroot.py
+++ b/salt/modules/chroot.py
@@ -56,7 +56,8 @@ def exist(name):
     '''
     dev = os.path.join(name, 'dev')
     proc = os.path.join(name, 'proc')
-    return all(os.path.isdir(i) for i in (name, dev, proc))
+    sys = os.path.join(name, 'sys')
+    return all(os.path.isdir(i) for i in (name, dev, proc, sys))
 
 
 def create(name):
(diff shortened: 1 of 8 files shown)


aplanas:
After talking with @meaksh we agree on:

* Accumulate the commits under a single PR
* Make the PR WIP (?)
* Merge the changes per week

meaksh:
@aplanas , for marking the PR as WIP, just add [WIP] at the beginning of the PR title. That's enough for us. Thanks!

aplanas:
Removing the WIP tag, ready for review.

salt/modules/chroot.py
```diff
@@ -50,24 +50,25 @@ def __virtual__():
         return (False, 'Module chroot requires the command chroot')
 
 
-def exist(name):
+def exist(root):
     '''
```
meaksh: I would prevent changing the name of the already accepted `name` parameter in all these methods from the `chmod` module. Since this is not a major Salt update but these changes will be included as a patch in our 2019.2.0 package, we need to be careful here.

Whoever is already using the `chmod` module from our 2019.2.0 package will have troubles if they were using kwargs when calling those methods.

aplanas: I very much agree, but this module was never released yet. The code is in Neon, and the patch for neon with this change is pending.

meaksh: After doing some checks, I can confirm we neither release the yomi patch yet in any of our products, so it would be safe to merge this. Thanks! :+1:

aplanas:
Thanks!

PR #166, opened 2019-07-31
Fix aptpkg systemd call (bsc#1143301)

dincamihai:
### What does this PR do?
Sets a `--description` for the systemd-run command called by aptpkg to avoid long description strings that would cause crashes.

Fixes:  [1143301 - aptpkg module causes systemd to crash in Ubuntu 18.04](https://bugzilla.suse.com/show_bug.cgi?id=1143301")

### Previous Behavior
See bug.

### New Behavior
Sets the description to `salt.modules.aptpkg`

### Tests written?

Yes

PR Upstream: https://github.com/saltstack/salt/pull/54077


dincamihai:
Thanks for reviewing!

PR #172, opened 2019-08-20
Implement network.fqdns module function (bsc#1134860)

ESiebigteroth:
### What does this PR do?
I changed the lacation of the fqdns grains to network.fqdns module
I created a parameter to disable the fqdns

https://bugzilla.suse.com/show_bug.cgi?id=1134860

### Previous Behavior
With the current implementation (in salt core grains), when there is a DNS misconfiguration, salt would get stuck/crash in unpredictable ways.

### New Behavior
add parameter to disable the fqdns grain and move the code to `network.fqdns` module 

### Tests written?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -25,8 +25,10 @@ import zlib
 from errno import EACCES, EPERM
 import datetime
 import warnings
+import salt.modules.network
 
 from multiprocessing.pool import ThreadPool
+from salt.utils.network import _get_interfaces
 
 # pylint: disable=import-error
 try:
@@ -83,6 +85,7 @@ __salt__ = {
     'cmd.run_all': salt.modules.cmdmod._run_all_quiet,
     'smbios.records': salt.modules.smbios.records,
     'smbios.get': salt.modules.smbios.get,
+    'network.fqdns': salt.modules.network.fqdns,
 }
 log = logging.getLogger(__name__)
 
@@ -1506,17 +1509,6 @@ def _linux_bin_exists(binary):
         return False
 
 
-def _get_interfaces():
-    '''
-    Provide a dict of the connected interfaces and their ip addresses
-    '''
-
-    global _INTERFACES
-    if not _INTERFACES:
-        _INTERFACES = salt.utils.network.interfaces()
-    return _INTERFACES
-
-
 def _parse_lsb_release():
     ret = {}
     try:
@@ -2201,51 +2193,7 @@ def fqdns():
     Return all known FQDNs for the system by enumerating all interfaces and
     then trying to reverse resolve them (excluding 'lo' interface).
     '''
-    # Provides:
-    # fqdns
-
-    grains = {}
-    fqdns = set()
-
-    def _lookup_fqdn(ip):
-        try:
-            name, aliaslist, addresslist = socket.gethostbyaddr(ip)
-            return [socket.getfqdn(name)] + [als for als in aliaslist if salt.utils.network.is_fqdn(als)]
-        except socket.herror as err:
-            if err.errno in (0, HOST_NOT_FOUND, NO_DATA):
-                # No FQDN for this IP address, so we don't need to know this all the time.
-                log.debug("Unable to resolve address %s: %s", ip, err)
-            else:
-                log.error(err_message, err)
-        except (socket.error, socket.gaierror, socket.timeout) as err:
-            log.error(err_message, err)
-
-    start = time.time()
-
-    addresses = salt.utils.network.ip_addrs(include_loopback=False, interface_data=_get_interfaces())
-    addresses.extend(salt.utils.network.ip_addrs6(include_loopback=False, interface_data=_get_interfaces()))
-    err_message = 'Exception during resolving address: %s'
-
-    # Create a ThreadPool to process the underlying calls to 'socket.gethostbyaddr' in parallel.
-    # This avoid blocking the execution when the "fqdn" is not defined for certains IP addresses, which was causing
-    # that "socket.timeout" was reached multiple times secuencially, blocking execution for several seconds.
-
-    try:
-       pool = ThreadPool(8)
-       results = pool.map(_lookup_fqdn, addresses)
-       pool.close()
-       pool.join()
-    except Exception as exc:
-       log.error("Exception while creating a ThreadPool for resolving FQDNs: %s", exc)
-
-    for item in results:
-        if item:
-            fqdns.update(item)
-
-    elapsed = time.time() - start
-    log.debug('Elapsed time getting FQDNs: {} seconds'.format(elapsed))
-
-    return {"fqdns": sorted(list(fqdns))}
+    return __salt__['network.fqdns']()
 
 
 def ip_fqdn():
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -27,7 +27,6 @@ import datetime
 import warnings
 import salt.modules.network
 
-from multiprocessing.pool import ThreadPool
 from salt.utils.network import _get_interfaces
 
 # pylint: disable=import-error
@@ -109,7 +108,6 @@ HAS_UNAME = True
 if not hasattr(os, 'uname'):
     HAS_UNAME = False
 
-_INTERFACES = {}
 
 # Possible value for h_errno defined in netdb.h
 HOST_NOT_FOUND = 1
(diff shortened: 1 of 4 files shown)


salt/grains/core.py
```diff
@@ -2201,51 +2191,10 @@ def fqdns():
...
-
-    for item in results:
-        if item:
-            fqdns.update(item)
-
-    elapsed = time.time() - start
-    log.debug('Elapsed time getting FQDNs: {} seconds'.format(elapsed))
-
-    return {"fqdns": sorted(list(fqdns))}
+    opt = 'localhost'
```
dincamihai: we probably need to find a better default here

dincamihai:
Please read https://chris.beams.io/posts/git-commit/ and adapt the commit messages.

salt/grains/core.py
dincamihai: @brejoc do you have an idea here?

salt/grains/core.py
```diff
@@ -2200,52 +2190,12 @@ def fqdns():
...
+    if __opts__.get('disablefqdns', False) == False:
+        opt = __salt__['network.fqdns']()
+    return opt
```
dincamihai: we need to import salt.modules.network and add:

```python
__salt__ = {
    ...
    'network.fqdns': salt.modules.network.fqdns
}
```

salt/grains/core.py
```diff
@@ -2200,52 +2192,12 @@ def fqdns():
...
-    for item in results:
-        if item:
-            fqdns.update(item)
-
-    elapsed = time.time() - start
-    log.debug('Elapsed time getting FQDNs: {} seconds'.format(elapsed))
-
-    return {"fqdns": sorted(list(fqdns))}
+    opt = {"fqdns": ['localhost']}
+    if __opts__.get('enablefqdns_grains', True) == True:
```
dincamihai: I think it is better to follow the same naming as here: https://github.com/openSUSE/salt/blob/199c520b1fff8b085fd8ee70406fb06237f1e42a/salt/grains/core.py#L204

dincamihai: so enable_fqdns_grains

salt/grains/core.py
dincamihai: Maybe this is helpful: https://serverfault.com/questions/269838/what-is-the-difference-between-a-hostname-and-a-fully-qualified-domain-name

salt/grains/core.py
```diff
@@ -2200,52 +2192,12 @@ def fqdns():
...
-
-    for item in results:
-        if item:
-            fqdns.update(item)
-
-    elapsed = time.time() - start
-    log.debug('Elapsed time getting FQDNs: {} seconds'.format(elapsed))
-
-    return {"fqdns": sorted(list(fqdns))}
+    opt = {"fqdns": ['localhost']}
```
dincamihai: Let's just return `{"fqdns": []}`

salt/modules/network.py
```diff
@@ -1881,3 +1885,59 @@ def iphexval(ip):
...
+    addresses = salt.utils.network.ip_addrs(include_loopback=False, interface_data=salt.utils.network._get_interfaces())
+    addresses.extend(salt.utils.network.ip_addrs6(include_loopback=False, interface_data=salt.utils.network._get_interfaces()))
+    err_message = 'Exception during resolving address: %s'
+
+    # Create a ThreadPool to process the underlying calls to 'socket.gethostbyaddr' in parallel.
+    # This avoid blocking the execution when the "fqdn" is not defined for certains IP addresses, which was causing
+    # that "socket.timeout" was reached multiple times secuencially, blocking execution for several seconds.
+
+    try:
+       pool = ThreadPool(8)
```
moio: Is retaining the multi-thread mechanism really needed? Running any command on the minion happens in an own process, so that shouldn't block other operations.

mcalmer: I think this is more about resolving names in parallel instead of sequential.

salt/modules/network.py
```diff
@@ -1881,3 +1885,59 @@ def iphexval(ip):
...
+       results = pool.map(_lookup_fqdn, addresses)
```
mcalmer: Does this really convert the list into a set and removes duplicates?

salt/modules/network.py
dincamihai: > Is retaining the multi-thread mechanism really needed? Running any command on the minion happens in an own process, so that shouldn't block other operations.

good point

dincamihai: let's wait for @meaksh (who implemented it) to conclude if we still need the parallel execution or not (he will be back on Monday)

meaksh: As @mcalmer mentioned here, the multi-thread mechanism is used to speedup the fqdns resolution, so we do it in parallel instead of sequentially. 

IMHO we should keep it even if we're now into an execution module, otherwise, in case of multiple IPs it would take longer.

salt/grains/core.py
```diff
@@ -25,8 +25,10 @@
 from errno import EACCES, EPERM
 import datetime
 import warnings
+import salt.modules.network
 
 from multiprocessing.pool import ThreadPool
```
meaksh: The `ThreadPool` import should be removed from here now.

salt/grains/core.py
```diff
@@ -1506,17 +1509,6 @@ def _linux_bin_exists(binary):
         return False
 
 
-def _get_interfaces():
-    '''
-    Provide a dict of the connected interfaces and their ip addresses
-    '''
-
-    global _INTERFACES
```
meaksh: The global `_INTERFACES` should be removed from this file: https://github.com/openSUSE/salt/blob/56701aebb0937251fb128b3e9a6ce914efb27e8f/salt/grains/core.py#L112

since we have it now in `salt.utils.network`:

https://github.com/openSUSE/salt/blob/56701aebb0937251fb128b3e9a6ce914efb27e8f/salt/utils/network.py#L58

ESiebigteroth: Hi Pablo I done your requested changes

salt/grains/core.py
```diff
@@ -2200,52 +2190,12 @@ def fqdns():
...
-    for item in results:
-        if item:
-            fqdns.update(item)
-
-    elapsed = time.time() - start
-    log.debug('Elapsed time getting FQDNs: {} seconds'.format(elapsed))
-
-    return {"fqdns": sorted(list(fqdns))}
+    opt = {"fqdns": []}
+    if __opts__.get('enable_fqdns_grains', True) == True:
```
meaksh: Do we really want to have the "fqdns" calculation in the grains by default on Salt, regardless of SUMA?

I would instead set the default behavior to `False` (not including fqdns in grains) and, in case this grain is still needed by SUMA, then explicitly add `enable_fqdns_grains` only on the SUSE Manager context.

@moio, @dincamihai - what do you think?

dincamihai: Isn't the current behavior to include the fqdns grains?

meaksh: Yes, that is the current behavior. But we want to get rid of that because the unpredictable problems when the network configuration is wrong and core grains takes too long into calculate.

dincamihai: From here: https://github.com/SUSE/spacewalk/issues/8428#issuecomment-517565935
The idea was that we don't change the default behavior with this PR but we allow SUSE Manager to disable the fqdns grains with a config param.

tests/unit/grains/test_core.py
```diff
@@ -845,6 +846,42 @@ def test_dns_return(self):
         with patch.object(salt.utils.dns, 'parse_resolv', MagicMock(return_value=resolv_mock)):
             assert core.dns() == ret
 
+
+    def test_enablefqdnsFalse(self):
+        '''
+        tests enable_fqdns_grains is set to False
+        '''
+        with patch.dict('salt.grains.core.__opts__', {'enable_fqdns_grains':False}):
+
```
meaksh: Just a nitpick: I would remove this empty line.

tests/unit/grains/test_core.py
```diff
@@ -845,6 +846,42 @@ def test_dns_return(self):
...
+            with patch.dict('salt.grains.core.__opts__', {'enable_fqdns_grains':True}):
+                assert core.fqdns() == 'my.fake.domain'
+
+
+    def test_enablefqdnsNone(self):
+        '''
+        testing default fqdns grains is returned when enable_fqdns_grains is None
+        '''
+        with patch.dict('salt.grains.core.__opts__', {'enable_fqdns_grains':None}):
+
```
meaksh: Remove empty line.

salt/grains/core.py
moio: Idea was to be nice at upstream users

dincamihai:
Included upstream in this PR: https://github.com/saltstack/salt/pull/52527

PR #177, opened 2019-08-30
Restore default behaviour of pkg list return

brejoc:
### What does this PR do?

The default behaviour for pkg list return was to not include patches,
even when installing patches. Only the packages where returned. There
is now parameter to also return patches if that is needed.

https://github.com/SUSE/spacewalk/issues/9245

### Previous Behavior

While installing patches also patches where in the return list.

### New Behavior

Only packages are returned now.

### Tests written?

No. Simply restored default behaviour.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1302,8 +1302,10 @@ def refresh_db(root=None):
     return ret
 
 
-def _find_types(pkgs):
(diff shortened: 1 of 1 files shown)


dincamihai:
Looks good. I think it will fix the P1 :)

aplanas:
Alternative in upstream: https://github.com/saltstack/salt/pull/54379

PR #185, opened 2019-11-05
virt.pool_running improvements

cbosdo:
### What does this PR do?

Enhances the `virt.pool_running` state to allow editing the pool and handle the `test` parameter.

backport of saltstack/salt#55202

### Tests written?

Yes


salt/modules/virt.py
```diff
@@ -683,21 +684,24 @@ def _gen_pool_xml(name,
...
+                  source_format=None,
+                  source_initiator=None):
     '''
     Generate the XML string to define a libvirt storage pool
     '''
     hosts = [host.split(':') for host in source_hosts or []]
     source = None
-    if source_devices or source_dir or source_adapter or hosts or source_auth or source_name or source_format:
+    if source_devices or source_dir or source_adapter or hosts or source_auth or source_name or source_format \
+            or source_initiator:
```
aplanas: Can I suggest to use `any()`?

salt/states/virt.py
```diff
@@ -788,6 +789,7 @@ def pool_running(name,
                                                  permissions=permissions,
                                                  source_devices=(source or {}).get('devices', None),
                                                  source_dir=(source or {}).get('dir', None),
+                                                 source_initiator=(source or {}).get('initiator', None),
```
aplanas: FYI `.get()` already return `None` if the element is missing, can save a lot of boilerplate.

```
>>> {}.get('key') is None
True
```

salt/modules/virt.py
cbosdo: Thanks for the tip, I didn't know `any()`. I'll update the upstream PR and this will cascaded a little later here

salt/states/virt.py
cbosdo: oh right

cbosdo:
@brejoc @meaksh The upstream equivalent PR has been merged. Could you merge this one?

brejoc:
Merged in ab972301f2b7488476b53c6a5b17b94ebaf22f78

PR #187, opened 2019-11-18
chroot: add apply_, sls and highstate for state execution

aplanas:
### What does this PR do?

Based on the code from the SSH client wrapper, add apply_, sls and highstate for the execution of Salt states inside a chroot environment.

### Tests written?

Yes

(backport of https://github.com/saltstack/salt/pull/55345)


aplanas:
this was already merged upstream some time ago

PR #190, opened 2019-11-26
Fix StreamClosed issue

dincamihai:
Fixes StreamClosed issue

Ref: https://github.com/SUSE/spacewalk/issues/10158

### DIFF
--- a/salt/cli/batch_async.py
+++ b/salt/cli/batch_async.py
@@ -221,7 +221,6 @@ class BatchAsync(object):
                 "metadata": self.metadata
             }
             self.event.fire_event(data, "salt/batch/{0}/done".format(self.batch_jid))
-            self.event.remove_event_handler(self.__event_handler)
             for (pattern, label) in self.patterns:
                 if label in ["ping_return", "batch_run"]:
                     self.event.unsubscribe(pattern, match_type='glob')
@@ -265,6 +264,7 @@ class BatchAsync(object):
 
     def __del__(self):
         self.local = None
+        self.event.remove_event_handler(self.__event_handler)
         self.event = None
         self.ioloop = None
         gc.collect()


salt/cli/batch_async.py
```diff
@@ -221,7 +221,6 @@ def end_batch(self):
                 "metadata": self.metadata
             }
             self.event.fire_event(data, "salt/batch/{0}/done".format(self.batch_jid))
-            self.event.remove_event_handler(self.__event_handler)
```
moio: @dincamihai curiosity: who/what used the event handler after `end_batch` and before `__del__`?

Bischoff:
Testing with that patch applied...

dincamihai:
> Tested with that patch applied to salt in server, proxy, and minion.
> 
> Sorry, but does not seem to work (the event stays stuck).

It seems salt-master was not restarted. Test need to be done again.

salt/cli/batch_async.py
dincamihai: something in thin call self.event.remove_event_handler(self.__event_handler)
so i've moved it in __del__

Bischoff:
> It seems salt-master was not restarted. Test need to be done again.

right, redoing properly.

salt/cli/batch_async.py
moio: Not sure I understand. Are you saying that `remove_event_handler` uses the event handler itself?

dincamihai: I have not actually investigated more after I've observed that moving that line fixed the problem.

What I think it happens is that we probably call `remove_event_handler` too soon.
It could be that when we call it, this `__del__` would be called: https://github.com/openSUSE/salt/blob/6c84612b52b5f14e74a1c44f03d78a85c6f0c5dc/salt/transport/ipc.py#L367 which would call https://github.com/openSUSE/salt/blob/6c84612b52b5f14e74a1c44f03d78a85c6f0c5dc/salt/transport/ipc.py#L384 which would probably trigger more cleanup.

I can see that `fire_event` should happen before `remove_event_handler`, but since `fire_event` is spawning a callback https://github.com/openSUSE/salt/blob/189a19b6e8d28cc49e5ad5f2a683e1dfdce66a86/salt/utils/event.py#L756 it could still be that `remove_event_handler` is called before `fire_event`. Maybe even before the `fire_event` for batch-start.

I know this explanation has some amount of assumptions, if you want me to be more pricese I would have to do more debugging :)

Let me know if it makes sense.

moio: It does make sufficient sense for the problem at hand.

The only thing I would recommend (only if you have some time) is to repeat the API stress test to make sure we aren't leaking any after this change.

dincamihai: Ok. I can do that. I will post back the outcome.

dincamihai: image

I executed the stress test with this patch and the memory leak is not present.
Will merge and promote on Monday.

moio: :+1:

PR #191, opened 2019-12-03
Let salt-ssh use 'platform-python' binary in RHEL8

cbbayburt:
RHEL/CentOS 8 has an internal Python interpreter called 'platform-python'
included in the base setup.
    
Add this binary to the list of Python executables to look for when creating the sh shim.
### Previous Behavior
In RHEL8, `salt-ssh` failed unless Python is installed from the Python module.

### New Behavior
`salt-ssh` works with the base `platform-python` binary.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -147,7 +147,7 @@ elif [ "$SUDO" ] && [ -n "$SUDO_USER" ]
 then SUDO="sudo "
 fi
 EX_PYTHON_INVALID={EX_THIN_PYTHON_INVALID}
-PYTHON_CMDS="python3 python27 python2.7 python26 python2.6 python2 python"
+PYTHON_CMDS="python3 /usr/libexec/platform-python python27 python2.7 python26 python2.6 python2 python"
 for py_cmd in $PYTHON_CMDS
 do
     if command -v "$py_cmd" >/dev/null 2>&1 && "$py_cmd" -c "import sys; sys.exit(not (sys.version_info >= (2, 6)));"


salt/client/ssh/__init__.py
```diff
@@ -147,7 +147,7 @@
 then SUDO="sudo "
 fi
 EX_PYTHON_INVALID={EX_THIN_PYTHON_INVALID}
-PYTHON_CMDS="python3 python27 python2.7 python26 python2.6 python2 python"
+PYTHON_CMDS="python3 /usr/libexec/platform-python3.6 python27 python2.7 python26 python2.6 python2 python"
```
meaksh: Just a minor question - why not doing it more generic? like:

```suggestion
PYTHON_CMDS="python3 /usr/libexec/platform-python python27 python2.7 python26 python2.6 python2 python"
```

Would that work?

aplanas: if that work, I would move it to the end of the list, so in case that another python package is installed, that will be used

mcalmer: well, python 3 should be prefered over python 2.
The code iterrate over the list and the first wins.

cbbayburt: There's also a `/usr/libexec/platform-python` in the package which links to `/usr/libexec/platform-python3.6`. I can add this to the end of the list as well, but I think it's not really necessary.

cbbayburt: @meaksh I wanted to explicitly add the version so the ordering is clear, but since it's not `3` but `3.6`, I think your suggestion makes sense to future-proof this.

PR #193, opened 2019-12-11
xfs: do not fails if type is not present

aplanas:
### What does this PR do?

The command `blkid -o export` not always provides a 'TYPE' output
for all the devices. One example is non-formatted partitions, like for
example the BIOS partition.

This patch do not force the presence of this field in the blkid
output.

(cherry picked from commit 88df6963470007aa4fe2adb09f000311f48226a8)

https://github.com/saltstack/salt/pull/53447

### Tests written?

Yes


dincamihai:
thanks @aplanas

PR #202, opened 2020-01-15
various netapi fixes and tests

brejoc:
### What does this PR do?

Various netapi fixes and tests.

https://github.com/saltstack/salt/pull/55865

### Previous Behavior

Authentication was not required when using the Salt API and specifying that the client type should be SSH. Additionally the ability to pass raw_shell commands with the Salt API was enabled by default, this allowed sending arbitrary commands to the SSH binary when using a client type of SSH.

### New Behavior

Authentication is required when using Salt API and a client type of SSH. The ability to use raw_shell commands is disabled by default.


brejoc:
Just for clarification: This basically is **the** change for 2019.2.3.

PR #203, opened 2020-01-16
Fix for Rhel8 integration test

brejoc:
### What does this PR do?

Dict structure seems to have changed. It's nested now and needs to be queried deeper.


brejoc:
I think that's not needed anymore.

PR #206, opened 2020-01-27
Delete bad API token files

brejoc:
### What does this PR do?

Under the following conditions and API token should be considered
invalid:

- The file is empty.
- We cannot deserialize the token from the file.
- The token exists but has no expiration date.
- The token exists but has expired.

All of these conditions necessitate deleting the token file. Otherwise
we should simply return an empty token.

https://github.com/saltstack/salt/pull/54388
bsc#1160931

### DIFF
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -25,7 +25,9 @@ from salt.ext import six
 
 # Import salt libs
 import salt.config
+import salt.exceptions
 import salt.loader
+import salt.payload
 import salt.transport.client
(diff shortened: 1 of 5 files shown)


brejoc:
@meaksh Okay, we've got something weird here. [This test](https://github.com/openSUSE/salt/blob/0331a32865e51ddc09d5cfee51d39dbb4e577d79/tests/unit/tokens/test_localfs.py) is missing completely from our repo and during cherr-pick it was marked as "deleted by us". :thinking:

Issue #213, opened 2020-02-19
Upstreaming patches: do not enforce zypper refresh if not needed


darix:
If someone can tell me how to make a master/minion option for "force" or "no force" I will happily extend the patch accordingly. but this patch reduced our highstate times a lot.

We setup the refresh option on the repositories that need it (updates- and buildservice repositories) and leave it out at the static GA repositories.

With the current code you can still force a force refresh via a salt module call.

```
Index: salt-2019.2.2-suse/salt/modules/zypperpkg.py
===================================================================
--- salt-2019.2.2-suse.orig/salt/modules/zypperpkg.py
+++ salt-2019.2.2-suse/salt/modules/zypperpkg.py
@@ -1277,9 +1277,10 @@ def mod_repo(repo, **kwargs):
     return repo
 
 
-def refresh_db(root=None):
+def refresh_db(root=None, force=False):
     '''
-    Force a repository refresh by calling ``zypper refresh --force``, return a dict::
+    Force a repository refresh by calling ``zypper refresh``, if the "force=True" flag is passed it will run with ``--force``.
+    It will return a dict::
 
         {'<database name>': Bool}
 
@@ -1290,12 +1291,15 @@ def refresh_db(root=None):
 
     .. code-block:: bash
 
-        salt '*' pkg.refresh_db
+        salt '*' pkg.refresh_db [force=true|false]
     '''
     # Remove rtag file to keep multiple refreshes from happening in pkg states
     salt.utils.pkg.clear_rtag(__opts__)
     ret = {}
-    out = __zypper__(root=root).refreshable.call('refresh', '--force')
+    refresh_opts = ['refresh']
+    if force:
+        refresh_opts.append('--force')
+    out = __zypper__(root=root).refreshable.call(*refresh_opts)
 
     for line in out.splitlines():
         if not line:
```

isbm:
Default was `force=True`. I would keep it that way. Second you probably don't want to have it in config, but probably just use pillar instead. That would be something like:

```yaml
zypper:
  refreshdb: false
```

```python
if __pillar__.get("zypper", {}).get("refreshdb", True):
    refresh_opt.append("--force")
```

brejoc:
I think this issue can be close. Please re-open if you think we still need it.

PR #215, opened 2020-02-20
Option to en-/disable force refresh in zypper

darix:
### What does this PR do?

Adds option to turn off force refresh

#213

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1277,9 +1277,10 @@ def mod_repo(repo, **kwargs):
     return repo
 
 
-def refresh_db(root=None):
+def refresh_db(root=None, force=None):
     '''
-    Force a repository refresh by calling ``zypper refresh --force``, return a dict::
+    Force a repository refresh by calling ``zypper refresh``, if the "force=True" flag is passed it will run with ``--force``.
+    It will return a dict::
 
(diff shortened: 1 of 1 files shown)


brejoc:
@darix Should we push this upstream too?

darix:
Anything blocking the merge?

brejoc:
@darix Sorry for the delay, but most of our resources where blocked by the Salt3000 preparations. We are currently opening PRs at upstream to get at least a weather forecast on the changes. Sometimes they'd like to see additional changes and then we'd have to update our patches again, because otherwise we'd diverge. So once we've got at least a :+1: from upstream, we can merge this and proceed.

agraul:
The upstream PR for this change can be found at https://github.com/saltstack/salt/pull/56419

brejoc:
Let's wait for them to give a thumbs up and then we can proceed here.

darix:
it is merged \o/

brejoc:
@agraul Did you check if we have https://github.com/saltstack/salt/pull/54954? Apart from that we are green here!

darix:
I have a patch locally that is on top of the root param patch.

osc rdiff openSUSE:{Factory,infrastructure}/salt

agraul:
@brejoc I just did a quick check, it looks like that PR is already covered.

PR #216, opened 2020-02-20
Returns a the list of IPs filtered by the optional network list.

### LINKED ISSUE #212: Upstreaming patches: filter_by_networks
```
Index: salt-2019.2.2-suse/salt/utils/network.py
===================================================================
--- salt-2019.2.2-suse.orig/salt/utils/network.py
+++ salt-2019.2.2-suse/salt/utils/network.py
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
 
     compliant = re.compile(r"(?!-)[A-Z\d\-\_]{1,63}(?<!-)$", re.IGNORECASE)
     return "." in hostname and len(hostname) < 0xff and all(compliant.match(x) for x in hostname.rstrip(".").split("."))
+
+
+def _filter_by_networks(values, networks):
+    filtered_addresses = []
+
+    for value in values:
+        address = ipaddress.ip_address(value)
+        for network in networks:
+            if network.__contains__(address):
+                filtered_addresses.append(value)
+
+    return filtered_addresses
+
+
+@jinja_filter('filter_by_networks')
+def filter_by_networks(values, filters=None):
+    '''
+    Returns a the list of IPs filtered by the optional network list.
+    If no filter is specified all all items are returned.
+    {% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
+        networks = [ipaddress.ip_network(network) for network in filters]
+        new_values = None
+        if isinstance(values, dict):
+            new_values = {}
+            for interface, addresses in values.items():
+                new_values[interface] = _filter_by_networks(addresses,
+                                                            networks)
+        elif isinstance(values, list):
+            new_values = _filter_by_networks(values, networks)
+        else:
+            raise ValueError('Do not know how to filter a {}'.format(type(values)))
+        return new_values
+    else:
+        return values
```

darix:
If no filter is specified all all items are returned.

```
{% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
{{ grains['ip_interfaces'] | filter_by_networks(networks) }}
{{ grains['ipv6'] | filter_by_networks(networks) }}
{{ grains['ipv4'] | filter_by_networks(networks) }}
```

Fixes #212

### DIFF
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -4,7 +4,9 @@ Define some generic socket functions for network modules
 '''
 
 # Import python libs
-from __future__ import absolute_import, unicode_literals, print_function
+from __future__ import absolute_import, print_function, unicode_literals
+
+import collections
 import itertools
 import os
 import re
@@ -14,6 +16,7 @@ import logging
 import platform
 import random
 import subprocess
+import collections
 from string import ascii_letters, digits
 
 # Import 3rd-party libs
@@ -2032,3 +2035,34 @@ def is_fqdn(hostname):
 
     compliant = re.compile(r"(?!-)[A-Z\d\-\_]{1,63}(?<!-)$", re.IGNORECASE)
     return "." in hostname and len(hostname) < 0xff and all(compliant.match(x) for x in hostname.rstrip(".").split("."))
+
+
+@jinja_filter("filter_by_networks")
+def filter_by_networks(values, networks):
+    """
+    Returns the list of IPs filtered by the network list.
+    If the network list is an empty sequence, no IPs are returned.
+    If the network list is None, all IPs are returned.
+
+    {% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    """
+
+    _filter = lambda ips, networks: [
+        ip for ip in ips for net in networks if ipaddress.ip_address(ip) in net
+    ]
+
+    if networks is not None:
+        networks = [ipaddress.ip_network(network) for network in networks]
+        if isinstance(values, collections.Mapping):
+            return {
+                interface: _filter(values[interface], networks) for interface in values
+            }
+        elif isinstance(values, collections.Sequence):
+            return _filter(values, networks)
+        else:
+            raise ValueError("Do not know how to filter a {}".format(type(values)))
+    else:
+        return values
--- a/tests/unit/utils/test_network.py
+++ b/tests/unit/utils/test_network.py
@@ -16,6 +16,8 @@ from tests.support.mock import (
     NO_MOCK,
     NO_MOCK_REASON,
 )
+import pytest
+import salt.exceptions
 
 # Import salt libs
 import salt.utils.network as network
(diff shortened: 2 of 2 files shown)


salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
 
     compliant = re.compile(r"(?!-)[A-Z\d\-\_]{1,63}(?<!-)$", re.IGNORECASE)
     return "." in hostname and len(hostname) < 0xff and all(compliant.match(x) for x in hostname.rstrip(".").split("."))
+
+
+def _filter_by_networks(values, networks):
```
isbm: I would still add a docstring here what does it do. :+1:

salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
...
+    filtered_addresses = []
+
+    for value in values:
+        address = ipaddress.ip_address(value)
+        for network in networks:
+            if network.__contains__(address):
```
isbm: Please don't do that. Pythonic way is just that:

```python
if address in network:
    ...
```

salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
...
+    If no filter is specified all all items are returned.
+    {% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
+        networks = [ipaddress.ip_network(network) for network in filters]
+        new_values = None
+        if isinstance(values, dict):
```
isbm: This is a bad idea. What if you get an OrderedMap instead? So the correct way is this:

```python
import collections

...

if isinstance(values, collections.Mapping):
    ...
```

salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
...
+            new_values = {}
+            for interface, addresses in values.items():
+                new_values[interface] = _filter_by_networks(addresses,
+                                                            networks)
+        elif isinstance(values, list):
```
isbm: This is also a bad idea for the same reasons:

```python
import collections

...

if isinstance(values, collections.Sequence):
   ...
```

salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
...
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
+        networks = [ipaddress.ip_network(network) for network in filters]
+        new_values = None
+        if isinstance(values, dict):
+            new_values = {}
+            for interface, addresses in values.items():
```
isbm: These days we do:

```python
for interface in values:
    new_values[interface] = _filter_by_networks(values[interface], networks)
```
Because it is memory-cheaper (JFYI: dict access is one of the cheapest operations in a Python bytecode) and is also faster.

salt/utils/network.py
```diff
@@ -2032,3 +2032,42 @@ def is_fqdn(hostname):
...
+def filter_by_networks(values, filters=None):
+    '''
+    Returns a the list of IPs filtered by the optional network list.
+    If no filter is specified all all items are returned.
+    {% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
```
isbm: Probably also:

```python

if filters and values:
```

brejoc:
@darix Can this be linked to a commit or an PR at upstream? Thx!

darix:
I was hoping you could upstream it afterwards? right now it is original work we need for our salt profiles at SUSE.

brejoc:
Ah, okay. At first the linked issue looked like this was a change that is already at upstream.

darix:
anything blocking the merge?

agraul:
I've opened https://github.com/saltstack/salt/pull/56394, once we get positive feedback from upstream we can merge this.

agraul:
I've pushed the commits that were merged upstream so we don't differ here. Can someone double check that I did it correctly pls @openSUSE/salt

meaksh:
Looks good to me @agraul . Thanks! :+1:

PR #217, opened 2020-02-26
batch_async: avoid using fnmatch to match event

moio:
### What does this PR do?

It reduces CPU load in cases where a lot of `async_batch` calls are being performed.

Behavior observed (`htop` output: before patch on top, after patch in bottom, load reduced to ~0.34x)

salt_pr_217

### Tests written?

No, already covered

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cli/batch_async.py
+++ b/salt/cli/batch_async.py
@@ -106,7 +106,7 @@ class BatchAsync(object):
             return
         mtag, data = self.event.unpack(raw, self.event.serial)
         for (pattern, op) in self.patterns:
-            if fnmatch.fnmatch(mtag, pattern):
+            if mtag.startswith(pattern[:-1]):
                 minion = data['id']
                 if op == 'ping_return':
                     self.minions.add(minion)


salt/minion.py
```diff
@@ -1443,6 +1443,11 @@ def _fire_master(self, data=None, tag=None, events=None, pretag=None, timeout=60
         else:
             return
 
+        if include_startup_grains:
+            grains_to_add = dict(
+                [(k, v) for k, v in six.iteritems(self.opts.get('grains', {})) if k in self.opts['start_event_grains']])
+            load['grains'] = grains_to_add
```
isbm: Much faster, more readable version that doesn't iterates every time on each grain, doesn't require new parameter and doesn't crashes if `start_event_grains` not there, but logs configuration typos:
```python
if "grains" in self.opts and bool(self.opts.get("start_event_grains")):
    load["grains"] = {}
    for k in self.opts["start_event_grains"]:
        try:
            load["grains"][k] = self.opts["grains"][k]
        except KeyError as exc:
            log.error("Cannot add event grain: {} not found".format(k))
```

salt/minion.py
```diff
@@ -2152,6 +2157,9 @@ def _refresh_grains_watcher(self, refresh_interval_in_minutes):
             })
 
     def _fire_master_minion_start(self):
+        include_grains = False
+        if self.opts['start_event_grains']:
+            include_grains = True
```
isbm: Actually just this:
```python
include_grains = bool(self.opts["start_event_grains"])
```
But this is not even needed, same as the whole new parameter to `_fire_master` is redundant and unnecessary.

moio:
@isbm wrong base, sorry, corrected now

moio:
@isbm can you re-review?

PR #218, opened 2020-03-02
use full option name instead of undocumented abbreviation

brejoc:
### What does this PR do?

Change an option name when call zypper. We used an undocumented abbreviation which will be removed in future versions of zypper. Use the full option name instead.

Prevents:

```
"Module function pkg.list_products threw an exception. Exception: Zypper command failure: The flag --disable-repos is not known.",
```

errors.

https://github.com/saltstack/salt/pull/56278


brejoc:
Test is needed, but this will be added later.

PR #220, opened 2020-03-10
Changed imports to vendored Tornado

brejoc:
### What does this PR do?

Switches to the vendored Tornado version where this was not yet the case or where it got reverted by patches from us.


isbm:
@meaksh @brejoc  There needs to be [a bit more work done](https://github.com/openSUSE/salt/blob/openSUSE-2019.2.0/salt/utils/thin.py#L25) than this. :wink:

brejoc:
@isbm Wrong branch. This fix is for v3000: https://github.com/openSUSE/salt/blob/openSUSE-3000/salt/utils/thin.py#L25

PR #221, opened 2020-03-12
loader: invalidate the import cachefor extra modules

aplanas:
Because we are mangling with importlib, we can find from time to
time an invalidation issue with sys.path_importer_cache, that
requires the removal of FileFinder that remain None for the
extra_module_dirs

(cherry picked from commit 0fb8e707a45d5caf40759e8b4943590d6fce5046)

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -1544,15 +1544,27 @@ class LazyLoader(salt.utils.lazy.LazyDict):
                 self._clean_module_dirs.append(directory)
 
     def __clean_sys_path(self):
+        invalidate_path_importer_cache = False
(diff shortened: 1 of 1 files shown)


brejoc:
@aplanas Is this already merged at upstream?

aplanas:
Is a new commit for this one: https://github.com/saltstack/salt/pull/53167

The rest of the batch is already in our package

brejoc:
@aplanas Do we also need this for 2019.23?

aplanas:
> @aplanas Do we also need this for 2019.2.3?

You are right. Yes.

PR #224, opened 2020-03-17
virt._get_domain: don't raise an exception if there is no VM

cbosdo:
Port of saltstack/salt#56392


meaksh:
I've already pushed this here at https://github.com/openSUSE/salt/commit/08886b6ff57b9e5e065c24fd4a4b5f3f004a5292

Thanks @cbosdo ! :+1:

PR #225, opened 2020-03-17
Fix pylint errors from PR #224

cbosdo:
Fix pylint errors squashed in saltstack/salt#56392


meaksh:
This change has been pushed and squashed here: ef376e2d9a8360367a9a214d8f50d56889f3a664 . Thanks!

PR #230, opened 2020-04-22
Catch NameError in IPCClient.close()

agraul:
### What does this PR do?
IPCClient's destructor calls close(), which tries to log a debug statement. Logging might hit a race condition, it sometimes raises:

    NameError: name '__salt_system_encoding__' is not defined

Logging not being available should not stop `__del__()` or `close()` from cleaning up.

Related issue: https://github.com/SUSE/spacewalk/issues/11018

### Previous Behavior

    NameError: name '__salt_system_encoding__' is not defined

### New Behavior

Exception is caught

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/transport/ipc.py
+++ b/salt/transport/ipc.py
@@ -361,8 +361,11 @@ class IPCClient(object):
 
         self._closing = True
 
-        log.debug('Closing %s instance', self.__class__.__name__)
-
+        try:
+            log.debug('Closing %s instance', self.__class__.__name__)
+        except NameError:
+            # Possible race condition: name '__salt_system_encoding__' is not defined
+            pass
         if self.stream is not None and not self.stream.closed():
             self.stream.close()
 


agraul:
Since products ship 2019.2.0, I am not sure if we need this in 2019.2.3 at all (but it should not hurt either).

meaksh:
@agraul we want to have this in Salt upstream. Could you please create the PR against Salt upstream `master` branch? Thanks!

agraul:
@meaksh I don't think it is needed in Salt-3000 and beyond - at least not for the exact error we ran into. Logging was changed and the `NameError` is caught in the logging module now:  https://github.com/agraul/salt/blob/master/salt/_logging/impl.py#L318

brejoc:
Right, this changed completely in Salt3000. So I think we are good to go.

agraul:
The PR was merged in 2019.2.0 and is not needed for Salt >= 3000 (it is caught somewhere else in newer versions). Since we don't ship 2019.2.3 we don't need to merge it here.

PR #234, opened 2020-05-07
msgpack: support versions >= 1.0.0

aplanas:
A recent change in msgpack >= 1.0.0, update the default value for the
parameter `raw` to False.  This change breaks Salt for those versions.

This patch add the parameter `raw=True` to all the unpack operations,
restoring the old default.

Fix bsc#1171257

(cherry picked from commit 1b3939fb01fc3405d8d222f118617220aecee092)

### DIFF
--- a/salt/utils/msgpack.py
+++ b/salt/utils/msgpack.py
@@ -69,12 +69,26 @@ def _sanitize_msgpack_kwargs(kwargs):
     return kwargs
 
 
+def _sanitize_msgpack_unpack_kwargs(kwargs):
+    """
+    Clean up msgpack keyword arguments for unpack operations, based on
+    the version
+    https://github.com/msgpack/msgpack-python/blob/master/ChangeLog.rst
+    """
+    assert isinstance(kwargs, dict)
+    if version >= (1, 0, 0) and kwargs.get("raw", None) is None:
+        log.info("adding `raw=True` argument to msgpack call")
+        kwargs["raw"] = True
+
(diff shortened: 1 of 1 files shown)


brejoc:
@aplanas Is this backwards compatible?

brejoc:
@aplanas Ah, sorry, I missed this: `if version >= (1, 0, 0)`
So I it is. Thx!

salt/utils/msgpack.py
```diff
@@ -69,12 +69,26 @@ def _sanitize_msgpack_kwargs(kwargs):
...
 
 
+def _sanitize_msgpack_unpack_kwargs(kwargs):
+    """
+    Clean up msgpack keyword arguments for unpack operations, based on
+    the version
+    https://github.com/msgpack/msgpack-python/blob/master/ChangeLog.rst
+    """
+    assert isinstance(kwargs, dict)
+    if version >= (1, 0, 0) and kwargs.get("raw", None) is None:
```
meaksh, line 79: Why not adding this new condition to the current logic inside `_sanitize_msgpack_kwargs` method?

aplanas: Because is not a valid parameter for the Packager, produces an exception. Another alternative is add a new boolean parameter to `_sanitize_msgpack_kwargs` to indicate is we are packaging or unpackaging, but is equivalent to this one.

meaksh: Thanks for clarifying!

PR #237, opened 2020-05-13
Add docker logout

agraul:
Docker logout works analog to login. It takes none, one or more registries as
arguments. If there are no arguments, all known (specified in pillar)
docker registries are logged out of. If arguments are present, they are
interpreted as a list of docker registries to log out of.

Backport of https://github.com/saltstack/salt/pull/56439
Related Issue: https://github.com/SUSE/spacewalk/issues/11017

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -1442,6 +1442,86 @@ def login(*registries):
     return ret
 
 
+def logout(*registries):
+    """
+    .. versionadded:: 3001
+
+    Performs a ``docker logout`` to remove the saved authentication details for
+    one or more configured repositories.
+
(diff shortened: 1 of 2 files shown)


salt/modules/dockermod.py
```diff
@@ -1442,6 +1442,92 @@ def login(*registries):
     return ret
 
 
+def logout(*registries):
+    '''
```
isbm, line 1446: JFYI: https://github.com/saltstack/salt/commit/0b2a5613b345f17339cb90e60b407199b3d26980

agraul: Thanks for the reminder. I'll run `black` over the change to make possible future backports easier.

PR #241, opened 2020-05-19
info_installed works without status attr now

brejoc:
### What does this PR do?

If 'status' was excluded via attr, info_installed was no longer able to detect if a package was installed or not. Now info_installed adds the 'status' for the 'lowpkg.info' request again.

Upstream PR: https://github.com/saltstack/salt/pull/57779

### Previous Behavior

Uninstalled packages with config (status==rc) where shown as installed.

### New Behavior

Uninstalled packages are no longer shown.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -2867,6 +2867,15 @@ def info_installed(*names, **kwargs):
     failhard = kwargs.pop('failhard', True)
     kwargs.pop('errors', None)                # Only for compatibility with RPM
     attr = kwargs.pop('attr', None)           # Package attributes to return
+
+    # status is needed to see if a package is installed. So we have to add it,
+    # even if it's excluded via attr parameter. Otherwise all packages are
+    # returned.
+    if attr:
+        attr_list = set(attr.split(','))
+        attr_list.add('status')
+        attr = ','.join(attr_list)
+
     all_versions = kwargs.pop('all_versions', False)  # This is for backward compatible structure only
 
     if kwargs:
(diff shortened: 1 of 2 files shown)


brejoc:
Upstream PR and tests are still missing.

brejoc:
If this PR get's merged, we can revert these two PR's:

https://github.com/uyuni-project/uyuni/pull/2231
https://github.com/SUSE/spacewalk/pull/11510

salt/modules/aptpkg.py
```diff
@@ -2867,6 +2867,16 @@ def info_installed(*names, **kwargs):
     failhard = kwargs.pop('failhard', True)
     kwargs.pop('errors', None)                # Only for compatibility with RPM
     attr = kwargs.pop('attr', None)           # Package attributes to return
+
+    # status is needed to see if a package is installed. So we have to add it,
+    # even if it's excluded via attr parameter. Otherwise all packages are
+    # returned.
+    if attr:
+        attr_list = attr.split(',')
```
agraul, line 2875: [nitpick] If the order of the attributes is not relevant, this could be a Set instead of a List (removes the next `if` statement)

brejoc: Good point. Not sure if it's really faster though. But let's change it.

tests/unit/modules/test_aptpkg.py
```diff
@@ -257,6 +257,23 @@ def test_info_installed(self):
...
+        call __salt__['lowpkg.info'] with the 'status' attribute even if the user
+        is not asking for it in 'attr'. Otherwise info_installed would not be able
+        to check if the package is installed and would return everything.
+
+        :return:
+        '''
+        with patch('salt.modules.aptpkg.__salt__', {'lowpkg.info': MagicMock(return_value=LOWPKG_INFO)}) as wget_lowpkg:
+            ret = aptpkg.info_installed('wget', attr='version')
+            calls = wget_lowpkg['lowpkg.info'].call_args_list.pop()
+            self.assertIn('status', calls.kwargs['attr'])
```
brejoc, line 274: Before there is any discussion around this weird way of checking this: `set()`, which is used in `info_installed()` is changing the order with every execution. That's why `assert_called_with()` can't be used here.

PR #245, opened 2020-05-27
Docker logout uses pillar instead of config.get

agraul:
### What does this PR do?
Storing docker credentials outside of the pillar was only introduced in Salt 3000. When docker.logout was backported to the 2019 branch this was not taken into account. This commit uses the pillar exclusively again.

### Previous Behavior
`salt * docker.logout` reported the following error:
> Passed invalid arguments to docker.logout: option() got an unexpected keyword argument 'wildcard '
### New Behavior
`salt * docker.logout`
Does not report errors and returns the expected response.

### Tests written?
No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -1477,22 +1477,24 @@ def logout(*registries):
     # this usecase (see https://github.com/docker/docker-py/issues/1091)
 
     # To logout of all known (to Salt) docker registries, they have to be collected first
-    registry_auth = __salt__["config.get"]("docker-registries", {})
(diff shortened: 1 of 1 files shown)


brejoc:
@agraul Do we also need an upstream PR for this?

agraul:
I don't think so, Salt 3000+ uses the `config.get` method and `docker.logout` was only added upstream recently. AFAIK they don't backport new features to their 2019.2 release, or do they?

brejoc:
@agraul Ah right. No, they don't. Tx!

PR #254, opened 2020-07-23
Fix __mount_device wrapper

aplanas:
Some recent change in Salt is now doing the right thing, and calling the
different states with separated args and kwargs.  This change trigger a
hidden bug in the __mount_device decorator, that expect those parameter
to be in kwargs, as is happening during the test.

This patch change the way that the wrapper inside the decorator search
for the name and device parameters, first looking into kwargs and later
in args if possible.  A new test is introduced to exercise both cases.

Fix #58012

(cherry picked from commit 2089645e2478751dc795127cfd14d0385c2e0899)

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/changelog/58012.fixed
+++ b/changelog/58012.fixed
@@ -1,0 +1,1 @@
+Fix btrfs state decorator, that produces exceptions when creating subvolumes.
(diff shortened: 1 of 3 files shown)


agraul:
[Upstream PR](https://github.com/saltstack/salt/pull/58013 ) was reviewed, approved an labeled "merge ready"

PR #267, opened 2020-09-25
zypperpkg: ignore retcode 104 for search()

aplanas:
Fix bsc#1159670

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -98,6 +98,8 @@ class _Zypper(object):
     }
 
     LOCK_EXIT_CODE = 7
+    NOT_FOUND_EXIT_CODE = 104
+
     XML_DIRECTIVES = ['-x', '--xmlout']
     # ZYPPER_LOCK is not affected by --root
     ZYPPER_LOCK = '/var/run/zypp.pid'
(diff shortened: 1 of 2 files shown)


salt/modules/zypperpkg.py
```diff
@@ -405,7 +405,9 @@ def _get_available_versions(self):
         Get available versions of the package.
         :return:
         '''
-        solvables = self.zypper.nolock.xml.call('se', '-xv', self.name).getElementsByTagName('solvable')
+        solvables = self.zypper.nolock.xml.call(
+            "se", "-xv", self.name, success_retcodes=[104]
```
isbm, line 409: You already calling XML, do you still need "x" option?

salt/modules/zypperpkg.py
```diff
@@ -983,7 +985,11 @@ def _is_match(pkgname):
         return False
 
     root = kwargs.get('root') or None
-    for node in __zypper__(root=root).xml.call('se', '-s', *targets).getElementsByTagName('solvable'):
+    for node in (
+        __zypper__(root=root)
+        .xml.call("se", "-s", *targets, success_retcodes=[104])
```
isbm, line 990: Would be awesome not to have a magic hardcodee but define it somewhere as a constant above.

salt/modules/zypperpkg.py
aplanas: Seems that is not needed. Removed!

salt/modules/zypperpkg.py
aplanas: Good idea, yes.

isbm (approved):

isbm:
👍 👍

PR #275, opened 2020-10-05
bsc#1176024: Fix file/directory user and group ownership containing UTF-8 characters

vzhestkov:
Fixing of UTF-8 support in user/group name containing UTF-8 characters in file module and state


salt/modules/file.py
```diff
@@ -344,7 +344,7 @@ def user_to_uid(user):
     try:
         if isinstance(user, int):
             return user
-        return pwd.getpwnam(user).pw_uid
+        return pwd.getpwnam(user.encode('utf-8')).pw_uid
```
isbm, line 347: You should use `salt.utils.stringutils` instead.

vzhestkov: Hi Bo.
Fixed in 384a726

isbm: Yep, now it is the right thing. Thanks!

PR #278, opened 2020-10-13
Fix for bsc#1102248 - psutil is broken and so Process is not working …

vzhestkov:
…on Python 3 as it is implemented

### What does this PR do?
Merges `ps` module from upstream

https://github.com/SUSE/spacewalk/issues/8470

### Previous Behavior
Traceback (most recent call last):
  File "/home/bo/work/salt/salt/cli/caller.py", line 212, in call
    ret['return'] = func(*args, **kwargs)
  File "/home/bo/work/salt/salt/modules/ps.py", line 156, in top
    for idx, (diff, process) in enumerate(reversed(sorted(usage))):
TypeError: unorderable types: Process() < Process()

### New Behavior
Expected output of `ps.top` call

### DIFF
--- a/salt/modules/ps.py
+++ b/salt/modules/ps.py
@@ -1,31 +1,33 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 A salt interface to psutil, a system and process library.
 See http://code.google.com/p/psutil.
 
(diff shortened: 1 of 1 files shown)


brejoc:
Note to self: Upstream PR not clear; better to link this PR in spec.

PR #284, opened 2020-10-19
Fix novendorchange option

mseidl:
There were lots of conflicts when cherry picking.... 

### What does this PR do?

Fixes the novendorchange option in zypper... 

Fixes: No issue number

### Previous Behavior
What happens with novendorchange is: the option of "--no-allow-change" is visible or not.  The issue is, in zypper vendor changes are not allowed by default. So, the the behavior between "zypper dup --no-allow-vendor-change" and "zypper dup" is nothing... 

### New Behavior
Now adds --allow-vendor-change 

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [ ] Tests written/updated

### Commits signed with GPG?
No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/master/topics/development/contributing.html) for best practices.

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1617,7 +1617,7 @@ def upgrade(refresh=True,
             dryrun=False,
             dist_upgrade=False,
             fromrepo=None,
-            novendorchange=False,
+            novendorchange=True,
             skip_verify=False,
             no_recommends=False,
             root=None,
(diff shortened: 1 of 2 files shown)


salt/modules/zypperpkg.py
```diff
@@ -1617,12 +1617,12 @@ def upgrade(refresh=True,
...
             dist_upgrade=False,
             fromrepo=None,
-            novendorchange=False,
+            novendorchange=True,
             skip_verify=False,
             no_recommends=False,
             root=None,
             **kwargs):  # pylint: disable=unused-argument
-    '''
+    """
```
brejoc, line 1625: I think you are mixing double and single quote here.

brejoc (changes requested):

salt/modules/zypperpkg.py
meaksh, line 1625: @mseidl this change (the double quotes) must be removed here since it's breaking the doc string.
```suggestion
    '''
```

meaksh (changes requested):

meaksh (approved):

PR #287, opened 2020-10-27
grains: master can read grains

aplanas:
### What does this PR do?

Seems that the master can read grains too, and the secure boot grain is only for root.

upstream: https://github.com/saltstack/salt/pull/58520

### DIFF
--- a/salt/grains/extra.py
+++ b/salt/grains/extra.py
@@ -94,8 +94,14 @@ def __secure_boot():
     enabled = False
     sboot = glob.glob("/sys/firmware/efi/vars/SecureBoot-*/data")
     if len(sboot) == 1:
-        with salt.utils.files.fopen(sboot[0], "rb") as fd:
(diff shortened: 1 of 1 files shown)


salt/grains/extra.py
```diff
@@ -94,8 +94,14 @@ def __secure_boot():
     enabled = False
     sboot = glob.glob("/sys/firmware/efi/vars/SecureBoot-*/data")
     if len(sboot) == 1:
-        with salt.utils.files.fopen(sboot[0], "rb") as fd:
-            enabled = fd.read()[-1:] == b"\x01"
+        # The minion is usually running as a privileged user, but is
+        # not the case for the master.  Seems that the master can algo
```
agraul, line 98: I found a small typo:

```suggestion
        # not the case for the master.  Seems that the master can also
```

aplanas: done

agraul (approved):
lgtm!

PR #290, opened 2020-11-05
Add "migrated" state and GPG key management functions

aplanas:
### What does this PR do?
Add functions in rpm_lowpkg for PGP key management, and a new migrated state in pkgrepo to migrate repositories. For now this state is only supported in SUSE systems, but the code is ready to be expanded

Fixes: #58782

- [X] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [X] Tests written/updated

Upstream: https://github.com/saltstack/salt/pull/58784

### DIFF
--- a/salt/modules/rpm_lowpkg.py
+++ b/salt/modules/rpm_lowpkg.py
@@ -823,3 +823,154 @@ def checksum(*paths, **kwargs):
                                                         python_shell=False))
 
     return ret
+
+
+def list_gpg_keys(info=False, root=None):
+    """Return the list of all the GPG keys stored in the RPM database
+
+    .. versionadded:: TBD
+
+    info
+       get the key information, returing a dictionary instead of a
+       list
+
+    root
+       use root as top level directory (default: "/")
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' lowpkg.list_gpg_keys
+        salt '*' lowpkg.list_gpg_keys info=True
+
+    """
+    cmd = ["rpm"]
+    if root:
+        cmd.extend(["--root", root])
+    cmd.extend(["-qa", "gpg-pubkey*"])
+    keys = __salt__["cmd.run_stdout"](cmd, python_shell=False).splitlines()
+    if info:
+        return {key: info_gpg_key(key, root=root) for key in keys}
+    else:
+        return keys
+
+
+def info_gpg_key(key, root=None):
+    """Return a dictionary with the information of a GPG key parsed
+
+    .. versionadded:: TBD
+
+    key
+       key identificatior
+
+    root
+       use root as top level directory (default: "/")
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' lowpkg.info_gpg_key gpg-pubkey-3dbdc284-53674dd4
+
+    """
+    cmd = ["rpm"]
+    if root:
+        cmd.extend(["--root", root])
+    cmd.extend(["-qi", key])
+    info = __salt__["cmd.run_stdout"](cmd, python_shell=False)
+
+    res = {}
+    # The parser algorithm is very ad-hoc.  Works under the
+    # expectation that all the fields are of the type "key: value" in
+    # a single line, except "Description", that will be composed of
+    # multiple lines.  Note that even if the official `rpm` makes this
+    # field the last one, other (like openSUSE) exted it with more
+    # fields.
+    in_description = False
+    description = []
+    for line in info.splitlines():
+        if line.startswith("Description"):
+            in_description = True
+        elif in_description:
+            description.append(line)
+            if line.startswith("-----END"):
+                res["Description"] = "\n".join(description)
+                in_description = False
+        elif line:
+            key, _, value = line.partition(":")
+            value = value.strip()
+            if "Date" in key:
+                try:
+                    value = datetime.datetime.strptime(
+                        value, "%a %d %b %Y %H:%M:%S %p %Z"
+                    )
+                except ValueError:
+                    pass
+            elif "Size" in key:
+                try:
+                    value = int(value)
+                except TypeError:
+                    pass
+            elif "(none)" in value:
+                value = None
+            res[key.strip()] = value
+    return res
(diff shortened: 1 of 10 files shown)


salt/modules/rpm_lowpkg.py
```diff
@@ -823,3 +823,150 @@ def checksum(*paths, **kwargs):
...
+
+    .. code-block:: bash
+
+        salt '*' lowpkg.info_gpg_key gpg-pubkey-3dbdc284-53674dd4
+    """
+    cmd = ["rpm"]
+    if root:
+        cmd.extend(["--root", root])
+    cmd.extend(["-qi", key])
+    info = __salt__["cmd.run_stdout"](cmd, python_shell=False)
```
mcalmer, line 879: Maybe using querytags could be an option to prevent parsing errors when the format change

salt/modules/rpm_lowpkg.py
```diff
@@ -823,3 +823,150 @@ def checksum(*paths, **kwargs):
...
+
+    .. code-block:: bash
+
+        salt '*' lowpkg.remove_gpg_key gpg-pubkey-3dbdc284-53674dd4
+    """
+    cmd = ["rpm"]
+    if root:
+        cmd.extend(["--root", root])
+    cmd.extend(["-e", key])
+    return __salt__["cmd.retcode"](cmd) == 0
```
mcalmer, line 972: Gpg keys can also be stored as a file. So it maybe not enough to remove it from db.

mcalmer (commented):
Did you check the apt implementation? Is it compatible? If i remember correctly there is also a state for it.

salt/modules/rpm_lowpkg.py
aplanas: For the info a single key, I do not think so, as we should need the full name.

For listing all the keys, I need to check. This can make the process more robust. I am using the documented process (man rpmkey), so I do not feel unconfortable of using `rpm -qa gpg-pubkey*` I will comeback with what I found.

salt/modules/rpm_lowpkg.py
aplanas: Do you have more info?

I see that SUMA, for example, is adding a cert, but IIUC I do not have a way of determining if this cert belong to a repo or not. So in Salt I am using only this method.

For Yomi I am complementing this with this: https://github.com/openSUSE/yomi/commit/50d63b987a87ad5105d4b0d84c872c79bf5b2699#diff-e8896cfc1c873fe1c7a57e4e4b4c5ed673373afec28d60092b021b694a290090R13, but maybe there is a better way?

aplanas:
> Did you check the apt implementation? Is it compatible? If i remember correctly there is also a state for it.

No need, is explicitly excluded here: https://github.com/openSUSE/salt/pull/290/files#diff-c95395f205338e42ffb381f9847ec5f3c0a9bbcf692a88f1fb7319077339e3f1R790

mcalmer:
> > Did you check the apt implementation? Is it compatible? If i remember correctly there is also a state for it.
> 
> No need, is explicitly excluded here: https://github.com/openSUSE/salt/pull/290/files#diff-c95395f205338e42ffb381f9847ec5f3c0a9bbcf692a88f1fb7319077339e3f1R790

This was not what I mean. But I checked and there is no state module to manage GPG keys.
But in aptpkg module there are functions:

modules/aptpkg.py:def get_repo_keys():
modules/aptpkg.py:def add_repo_key(path=None, text=None, keyserver=None, keyid=None, saltenv="base"):
modules/aptpkg.py:def del_repo_key(name=None, **kwargs):

Maybe a good idea to make something compatible ?

salt/modules/rpm_lowpkg.py
mcalmer: Check `/usr/lib/rpm/gnupg/keys/` directory on your system. AFAIK RedHat is only using this dir.
Please check with our zypp experts. They know all the details.

aplanas:
> > > Did you check the apt implementation? Is it compatible? If i remember correctly there is also a state for it.
> > 
> > 
> > No need, is explicitly excluded here: https://github.com/openSUSE/salt/pull/290/files#diff-c95395f205338e42ffb381f9847ec5f3c0a9bbcf692a88f1fb7319077339e3f1R790
> 
> This was not what I mean. But I checked and there is no state module to manage GPG keys.
> But in aptpkg module there are functions:
> 
> modules/aptpkg.py:def get_repo_keys():
> modules/aptpkg.py:def add_repo_key(path=None, text=None, keyserver=None, keyid=None, saltenv="base"):
> modules/aptpkg.py:def del_repo_key(name=None, **kwargs):
> 
> Maybe a good idea to make something compatible ?

Indeed, I would expect to find them in `dpkg_lowpkg.py`. Is unfortunate that for rpm OS it should live in `rpm_lowpkg`, and this will imply to replicate it in all the `zypperpkg.py` and in `yumpkg.py` the same API to accommodate the one form Debian.

I can create this compatibility layer.

isbm:
@aplanas I've designed RFC about interfaces years go exactly for that reason: while `pkg install` would abstract OS, `pkg <specific-per-os-api>` completely breaks the whole idea. However, that was never accepted, maybe you could succeed trying it again.

salt/modules/rpm_lowpkg.py
aplanas: Uhm thanks. My testings shows that this is the directory where the rpm keyring is stored. Every import store there an .asc file, and a manual remove of a file, removes the key in rpm.

Seems to me that the documented way of removing (-e) and querying (-qi) are covering the use case, and are equivalent of manual adding and removing .asc files.

mcalmer: Could be. I was unsure if RPM is so good to keep its DB and this dir in sync or not:-)

aplanas:
> Maybe a good idea to make something compatible ?

It should be done in the last commit here and upstream

aplanas:
@mcalmer one last review?

mcalmer (approved):

Issue #296, opened 2020-11-24
numeric uid owner breaks salt file module/state


darix:
### Description of Issue

This output is from a state.highstate. the file on the machine is owned by an unresolvable UID. 

```
          ID: zypp_conf_solver_dupAllowVendorChange
    Function: file.replace
        Name: /etc/zypp/zypp.conf
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python3.6/site-packages/salt/state.py", line 1987, in call
                  ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
                File "/usr/lib/python3.6/site-packages/salt/loader.py", line 2031, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3.6/site-packages/salt/states/file.py", line 4823, in replace
                  backslash_literal=backslash_literal)
                File "/usr/lib/python3.6/site-packages/salt/modules/file.py", line 2485, in replace
                  check_perms(path, None, pre_user, pre_group, pre_mode)
                File "/usr/lib/python3.6/site-packages/salt/modules/file.py", line 4577, in check_perms
                  not salt.utils.platform.is_windows() and salt.utils.stringutils.to_str(user) != perms['luser']
                File "/usr/lib/python3.6/site-packages/salt/utils/stringutils.py", line 103, in to_str
                  raise TypeError('expected str, bytes, or bytearray not {}'.format(type(s)))
              TypeError: expected str, bytes, or bytearray not <class 'int'>
     Started: 16:44:07.373835
    Duration: 5.901 ms
     Changes:   
```

you can also reproduce it with

```
salt '*' file.check_perms /etc/zypp/locks '{}' 65433 root 644

hostname:
    Passed invalid arguments to file.check_perms: expected str, bytes, or bytearray not <class 'int'>
```

### Versions Report

```
Salt Version:
           Salt: 3000.3
 
Dependency Versions:
           cffi: 1.14.3
       cherrypy: Not Installed
       dateutil: 2.8.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.11.2
        libgit2: Not Installed
       M2Crypto: 0.36.0
           Mako: 1.1.3
   msgpack-pure: Not Installed
 msgpack-python: 1.0.0
   mysql-python: Not Installed
      pycparser: 2.20
       pycrypto: 3.9.8
   pycryptodome: Not Installed
[...]

PR #297, opened 2020-11-25
Fix salt.utils.stringutils.to_str calls to make it working with numeric uid/gid

vzhestkov:
### What does this PR do?
Fixes `salt.utils.stringutils.to_str` with `int` argument instead of expected str, bytes, or bytearray

https://github.com/openSUSE/salt/issues/296

### Previous Behavior
Backtrace on calling `file.check_perm` with numeric `uid` or `gid`.

### New Behavior
Works as expected with numeric `uid` or `gid` and with user and group names in latin and non latin characters.

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -4567,6 +4567,12 @@ def check_perms(name, ret, user, group, mode, attrs=None, follow_symlinks=False)
     is_dir = os.path.isdir(name)
     is_link = os.path.islink(name)
 
+    def __safe_to_str(s):
+        try:
+            return salt.utils.stringutils.to_str(s)
+        except:
+            return salt.utils.stringutils.to_str(str(s))
+
     # user/group changes if needed, then check if it worked
     if user:
         if isinstance(user, int):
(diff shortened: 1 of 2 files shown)


aplanas:
I would recommend to extend `stringutils.to_str` to default to `str()` if the type is not recognized.

vzhestkov:
I prefer this solution to change `stringutils.to_str`, but we have discussed it with @meaksh and I agreed that it could be unsafe for others who is using `stringutils.to_str`.

meaksh:
> I would recommend to extend `stringutils.to_str` to default to `str()` if the type is not recognized.

Hey @aplanas, we think that there might be logic in some other places in the Salt codebase (even external modules) that are relying on having a exception raised in some cases, if we prevent the exception within `stringutils.to_str` then this could lead into unexpected behaviors. So we agreed that it's safer to fix this in the file modules.

aplanas:
> Hey @aplanas, we think that there might be logic in some other places in the Salt codebase (even external modules) that are relying on having a exception raised in some cases

Ahm, I see. Thanks!

PR #304, opened 2020-12-09
Force zyppnotify to prefer Packages.db than Packages if it exists

vzhestkov:
### What does this PR do?
Some of the installed systems will use NDB backend DB instead of Berkeley DB.
For such case we should use `Packages.db` file instead of `Packages` to create cookie file.

[PM-1185](https://github.com/SUSE/spacewalk/issues/13159): RPM database changing

### Previous Behavior
`zyppnotify` is creating cookie based on `Packages` file.

### New Behavior
`zyppnotify` is creating cookie based on `Packages.db` file if it exists, otherwise will use `Packages` as it was before.

### DIFF
--- a/scripts/suse/zypper/plugins/commit/zyppnotify
+++ b/scripts/suse/zypper/plugins/commit/zyppnotify
@@ -19,7 +19,9 @@ class DriftDetector(Plugin):
     def __init__(self):
         Plugin.__init__(self)
         self.ck_path = "/var/cache/salt/minion/rpmdb.cookie"
-        self.rpm_path = "/var/lib/rpm/Packages"
+        self.rpm_path = "/var/lib/rpm/Packages.db"
+        if not os.path.exists(self.rpm_path):
+            self.rpm_path = "/var/lib/rpm/Packages"
 
     def _get_mtime(self):
         '''


scripts/suse/zypper/plugins/commit/zyppnotify
```diff
@@ -19,7 +19,9 @@ class DriftDetector(Plugin):
     def __init__(self):
         Plugin.__init__(self)
         self.ck_path = "/var/cache/salt/minion/rpmdb.cookie"
-        self.rpm_path = "/var/lib/rpm/Packages"
+        self.rpm_path = "/var/lib/rpm/Packages.db"
```
aplanas, line 22: later some methods read from this file in specific places, is this ndb compatible?

vzhestkov: Packages.db in this case is ndb file, it should be preferred one if exists. I just discussed it with @mlschroe.

aplanas: Indeed, but as https://github.com/openSUSE/salt/pull/304/files/42724735d506be32882f3d91a22f456354ef3a8a#diff-de57eeeae39d1af47a8c4bbbac040ff8fc6a38bab688dc1f72f755a148902cf8R43 we read a section from there, but I guess that is not binary compatible both formats?

vzhestkov: I see what you mean, but zyppnotify is reading not just one section of 0x1000 bytes from the file, it is reading it whole by blocks with size equal to 4096 and the cookie file contains sha256 hash of whole file.

aplanas: Ouch. Indeed, yes!

PR #313, opened 2021-02-02
add patch support for allow vendor change option with zypper

mseidl:
### What does this PR do?
Adds patch support for vendor change
But this fixes the version support for upgrades.  

Dist upgrades support vendor change with  sle12+ 
Install/patch/upgrade support vendor change with sle15+

https://github.com/SUSE/spacewalk/pull/13370
https://github.com/openSUSE/zypp-plugin-spacewalk/pull/44

Also this can be merged without  the other's merged. 

### Previous Behavior

### New Behavior

### Tests written?
Please review the [test documentation](https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html) for details on how to implement tests into Salt's test suite.

Not yet

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -222,21 +222,20 @@ class _Zypper(object):
     def pid(self):
         return self.__call_result.get('pid', '')
 
-    def calc_zypper(self):
+    def refresh_zypper_flags(self):
         try:
-            zypp_tmp = version('zypper')
+            zypp_version = version('zypper')
             # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
-            if version_cmp(zypp_tmp, '1.11.34') >= 0 and version_cmp(zypp_tmp, '1.14.8') == -1:
+            if version_cmp(zypp_version, '1.11.34') >= 0:
                 # zypper version supports vendor change for dist upgrade
                 self.dup_avc = True
             # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
-            if version_cmp(zypp_tmp, '1.14.8') >= 0:
+            if version_cmp(zypp_version, '1.14.8') >= 0:
                 self.inst_avc = True
             else:
                 log.error("Failed to compare Zypper version")
         except Exception as ex:
-            log.error("Unable to get Zypper version")
-            log.error(ex)
+            log.error("Unable to get Zypper version: {}".format(ex))
 
     def _is_error(self):
         '''
@@ -1435,7 +1434,6 @@ def install(name=None,
     skip_verify
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
 
-
     novendorchange
         Disallow vendor change
 
@@ -1590,7 +1588,7 @@ def install(name=None,
     cmd_install.append(kwargs.get('resolve_capabilities') and '--capability' or '--name')
     # Install / patching / upgrade with vendor change support is only in SLE 15+  opensuse Leap 15+
     if not novendorchange:
-        __zypper__(root=root).calc_zypper()
+        __zypper__(root=root).refresh_zypper_flags()
         if __zypper__(root=root).inst_avc:
             cmd_install.append("--allow-vendor-change")
             log.info("Enabling vendor changes")
@@ -1608,6 +1606,7 @@ def install(name=None,
         cmd_install.extend(fromrepoopt)
     if no_recommends:
         cmd_install.append('--no-recommends')
+
     errors = []
 
     # Split the targets into batches of 500 packages each, so that
@@ -1733,11 +1732,10 @@ def upgrade(refresh=True,
             cmd_update.extend(['--from' if dist_upgrade else '--repo', repo])
         log.info('Targeting repos: %s', fromrepo)
 
-    # TODO: Grains validation should be moved to Zypper class
     if not novendorchange:
-        __zypper__(root=root).calc_zypper()
+        __zypper__(root=root).refresh_zypper_flags()
         if dist_upgrade:
-            if __zypper__(root=root).inst_avc or __zypper__(root=root).dup_avc:
+            if __zypper__(root=root).dup_avc:
                 cmd_update.append("--allow-vendor-change")
                 log.info("Enabling vendor changes")
             else:
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -105,6 +105,10 @@ class _Zypper(object):
     ZYPPER_LOCK = '/var/run/zypp.pid'
     TAG_RELEASED = 'zypper/released'
     TAG_BLOCKED = 'zypper/blocked'
+    # Dist upgrade vendor change support (SLE12+)
+    dup_avc = False
+    # Install/Patch/Upgrade vendor change support (SLE15+)
+    inst_avc = False
 
     def __init__(self):
         '''
@@ -217,6 +221,22 @@ class _Zypper(object):
     @property
     def pid(self):
         return self.__call_result.get('pid', '')
+
+    def calc_zypper(self):
+        try:
+            zypp_tmp = version('zypper')
+            # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
+            if version_cmp(zypp_tmp, '1.11.34') >= 0 and version_cmp(zypp_tmp, '1.14.8') == -1:
+                # zypper version supports vendor change for dist upgrade
+                self.dup_avc = True
+            # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
+            if version_cmp(zypp_tmp, '1.14.8') >= 0:
+                self.inst_avc = True
+            else:
+                log.error("Failed to compare Zypper version")
+        except Exception as ex:
+            log.error("Unable to get Zypper version")
+            log.error(ex)
 
     def _is_error(self):
         '''
@@ -1368,6 +1388,7 @@ def install(name=None,
             no_recommends=False,
             root=None,
             inclusion_detection=False,
+            novendorchange=True,
             **kwargs):
     '''
     .. versionchanged:: 2015.8.12,2016.3.3,2016.11.0
@@ -1413,6 +1434,10 @@ def install(name=None,
 
     skip_verify
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
+
+
+    novendorchange
+        Disallow vendor change
 
     version
         Can be either a version number, or the combination of a comparison
@@ -1563,6 +1588,15 @@ def install(name=None,
     cmd_install = ['install', '--auto-agree-with-licenses']
 
     cmd_install.append(kwargs.get('resolve_capabilities') and '--capability' or '--name')
+    # Install / patching / upgrade with vendor change support is only in SLE 15+  opensuse Leap 15+
+    if not novendorchange:
+        __zypper__(root=root).calc_zypper()
+        if __zypper__(root=root).inst_avc:
+            cmd_install.append("--allow-vendor-change")
+            log.info("Enabling vendor changes")
+        else:
+            log.warning("Enabling/Disabling vendor changes is not supported on this Zypper version")
+
 
     if not refresh:
         cmd_install.insert(0, '--no-refresh')
@@ -1574,7 +1608,6 @@ def install(name=None,
         cmd_install.extend(fromrepoopt)
     if no_recommends:
         cmd_install.append('--no-recommends')
-
     errors = []
 
     # Split the targets into batches of 500 packages each, so that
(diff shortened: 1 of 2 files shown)


salt/modules/zypperpkg.py
```diff
@@ -1700,28 +1718,41 @@ def upgrade(refresh=True,
...
-        # TODO: Grains validation should be moved to Zypper class
-        if __grains__["osrelease_info"][0] > 11:
-            if novendorchange:
-                cmd_update.append("--no-allow-vendor-change")
-                log.info("Disabling vendor changes")
+    # TODO: Grains validation should be moved to Zypper class
+    if novendorchange:
+        if dist_upgrade:
+            if __grains__["osrelease_info"][0] > 11:
+                if novendorchange:
```
brejoc, line 1725: I'm not sure if this makes sense. `novendorchange' is the default for SLE if I'm not mistaken. So if you set this to `False` we don't check if this is a dist-upgrade and also won't check if `novendorchange` is `False` to allow the vendor change then. Did I miss something?

brejoc (changes requested):
Not sure, but I think there might be one minor and one bigger problem. Let's talk about this maybe?

salt/modules/zypperpkg.py
```diff
@@ -1563,6 +1568,20 @@ def install(name=None,
     cmd_install = ['install', '--auto-agree-with-licenses']
 
     cmd_install.append(kwargs.get('resolve_capabilities') and '--capability' or '--name')
+    # Install / patching / upgrade with vendor change support is only in SLE 15+  opensuse Leap 15+
+    if novendorchange:
+        if __grains__["osrelease_info"][0] > 12:
+            cmd_install.append("--no-allow-vendor-change")
+            log.info("Disabling vendor changes")
+        else:
+            log.warning("Enabling/Disabling vendor changes is not supported on this Zypper version")
```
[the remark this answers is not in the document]
mseidl: Vendor change is supported with dist upgrades starting with  SLE12 SP0 (an update, not GA).  
Install/patching/upgrade with vendor change was not supported till SLE15 SP0 (an update, not GA).

salt/modules/zypperpkg.py
```diff
@@ -218,6 +222,22 @@ def stderr(self):
     def pid(self):
         return self.__call_result.get('pid', '')
 
+    def calc_zypper(self):
+        try:
+            zypp_tmp = version('zypper')
+            # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
+            if version_cmp(zypp_tmp, '1.11.34') >= 0 and version_cmp(zypp_tmp, '1.14.8') == -1:
```
meaksh, line 229: This code is doing some weird things with the flags.

For example, let assume the Zypper version is `1.15`. It should support "vendorchange" flags for everything (dup, install, patch..), but your flags would be set as:

```
dup_avc = False
inst_avc = True
```

And this is kind of inconsistent. Both flags should be `True` in that case. I'm not sure if I'm missing something.

I would remove this part of the version checking, to have consistency between the two flags

```suggestion
            if version_cmp(zypp_tmp, '1.11.34') >= 0:
```

Does that make sense for you?

salt/modules/zypperpkg.py
```diff
@@ -1700,28 +1733,35 @@ def upgrade(refresh=True,
...
         log.info('Targeting repos: %s', fromrepo)
 
-    if dist_upgrade:
-        # TODO: Grains validation should be moved to Zypper class
-        if __grains__["osrelease_info"][0] > 11:
-            if novendorchange:
-                cmd_update.append("--no-allow-vendor-change")
-                log.info("Disabling vendor changes")
-            else:
+    # TODO: Grains validation should be moved to Zypper class
```
meaksh, line 1736: There is no more grains verification. I think we can remove this comment now.
```suggestion
```

salt/modules/zypperpkg.py
```diff
@@ -1700,28 +1733,35 @@ def upgrade(refresh=True,
...
+    if not novendorchange:
+        __zypper__(root=root).calc_zypper()
+        if dist_upgrade:
+            if __zypper__(root=root).inst_avc or __zypper__(root=root).dup_avc:
```
meaksh, line 1740: In case you fixed the inconsistency with the `inst_avc` and `dup_avc` flags I mentioned above, I think this can be changed to:

```suggestion
            if __zypper__(root=root).dup_avc:
```

salt/modules/zypperpkg.py
mseidl: Yes, this is good thanks.

salt/modules/zypperpkg.py
```diff
@@ -218,6 +222,22 @@ def stderr(self):
...
+                # zypper version supports vendor change for dist upgrade
+                self.dup_avc = True
+            # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
+            if version_cmp(zypp_tmp, '1.14.8') >= 0:
+                self.inst_avc = True
+            else:
+                log.error("Failed to compare Zypper version")
+        except Exception as ex:
+            log.error("Unable to get Zypper version")
+            log.error(ex)
```
meaksh, line 239: This could be unified as:

```suggestion
            log.error("Unable to get Zypper version: {}".format(ex))
```

salt/modules/zypperpkg.py
```diff
@@ -218,6 +222,22 @@ def stderr(self):
...
 
+    def calc_zypper(self):
+        try:
+            zypp_tmp = version('zypper')
+            # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
+            if version_cmp(zypp_tmp, '1.11.34') >= 0:
+                # zypper version supports vendor change for dist upgrade
+                self.dup_avc = True
+            # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
+            if version_cmp(zypp_tmp, '1.14.8') >= 0:
```
meaksh, line 233: I think we should use a better naming here:

```suggestion
    def refresh_zypper_flags(self):
        try:
            zypp_version = version('zypper')
            # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
            if version_cmp(zypp_version, '1.11.34') >= 0:
                # zypper version supports vendor change for dist upgrade
                self.dup_avc = True
            # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
            if version_cmp(zypp_version, '1.14.8') >= 0:
```

salt/modules/zypperpkg.py
```diff
@@ -1563,6 +1588,15 @@ def install(name=None,
     cmd_install = ['install', '--auto-agree-with-licenses']
 
     cmd_install.append(kwargs.get('resolve_capabilities') and '--capability' or '--name')
+    # Install / patching / upgrade with vendor change support is only in SLE 15+  opensuse Leap 15+
+    if not novendorchange:
+        __zypper__(root=root).calc_zypper()
```
meaksh, line 1593:
```suggestion
        __zypper__(root=root).refresh_zypper_flags()
```

salt/modules/zypperpkg.py
```diff
@@ -1700,28 +1733,34 @@ def upgrade(refresh=True,
...
 
-    if dist_upgrade:
-        # TODO: Grains validation should be moved to Zypper class
-        if __grains__["osrelease_info"][0] > 11:
-            if novendorchange:
-                cmd_update.append("--no-allow-vendor-change")
-                log.info("Disabling vendor changes")
-            else:
+    if not novendorchange:
+        __zypper__(root=root).calc_zypper()
```
meaksh, line 1737:
```suggestion
        __zypper__(root=root).refresh_zypper_flags()
```

meaksh (commented):
Added few more comments.

salt/modules/zypperpkg.py
```diff
@@ -1414,6 +1434,10 @@ def install(name=None,
     skip_verify
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
 
+
```
meaksh, line 1437:
```suggestion

```

salt/modules/zypperpkg.py
```diff
@@ -1574,7 +1607,6 @@ def install(name=None,
         cmd_install.extend(fromrepoopt)
     if no_recommends:
         cmd_install.append('--no-recommends')
-
     errors = []
```
meaksh, line 1610:
```suggestion

    errors = []
```

meaksh (commented):
Added two minor nitpick. Otherwise looks good

meaksh (approved):

salt/modules/zypperpkg.py
```diff
@@ -218,6 +222,21 @@ def stderr(self):
...
+    def refresh_zypper_flags(self):
+        try:
+            zypp_version = version('zypper')
+            # zypper version 1.11.34 in SLE12 update supports vendor change for only dist upgrade
+            if version_cmp(zypp_version, '1.11.34') >= 0:
+                # zypper version supports vendor change for dist upgrade
+                self.dup_avc = True
+            # zypper version 1.14.8 in SLE15 update supports vendor change in install/patch/upgrading
+            if version_cmp(zypp_version, '1.14.8') >= 0:
+                self.inst_avc = True
```
isbm, line 234:
```python
self.dup_avc = version_cmp(zypp_version, "1.11.34") > -1
self.inst_avc = version_cmp(zypp_version, "1.14.8") > -1
```

Still needs refactoring on better version-to-feature tracking.

salt/modules/zypperpkg.py
```diff
@@ -1563,6 +1586,15 @@ def install(name=None,
     cmd_install = ['install', '--auto-agree-with-licenses']
 
     cmd_install.append(kwargs.get('resolve_capabilities') and '--capability' or '--name')
+    # Install / patching / upgrade with vendor change support is only in SLE 15+  opensuse Leap 15+
+    if not novendorchange:
+        __zypper__(root=root).refresh_zypper_flags()
+        if __zypper__(root=root).inst_avc:
+            cmd_install.append("--allow-vendor-change")
```
isbm, line 1593: BTW, this though wasn't the intention of the original design. This should happen instead:

```python
__zypper__(root=root).allow_vendor_change(novendorchange). ....
```

So all the logic of the vendor change-or-not should be in the Zypper's class above instead.

salt/modules/zypperpkg.py
brejoc: @isbm I'm not sure if I'm getting you proposal here. Why is this the suggested change improving the "version-to-feature tracking"? Or is this unrelated?

isbm: Change is only makes less codelines.

The problem I am referring to, is that you still have scattered hard-coded versions (and these will grow up), mapped to... nothing. So you will end up with "version x.y.z does this, that and that", "version  x.y.y does only this, but not that" etc etc. You don't want that. @meaksh  had a good idea with capabilities and I wrote in 2018 an RFC and implementation of remote inspection so you can construct Jinja the way it "knows" if a particular feature is there or not. Because bare Salt versions are not that helpful if you patch faster then rolling the releases (which is usually the case).

There is another problem you should solve: "Can client alice.acme.com perform feature X and bob.acme.com not?" — and before you even go there. Otherwise you will often end up with a semi-failed-but-still-good states, which you will still need to accept: something that e.g. Ansible does not have, and it is very good.

So for now you could do as-is: hardcoding versions. But in a long run the problem is still not solved.

@meaksh your time. :)

meaksh: Hey @isbm :)
I see we're pointing two different problems here:

1) Make `zypperpkg` to properly support `allow-vendor-change` flag, which are missing or broken atm (which is the purpose of this PR)
2) Determine the capabilties of a certain minion in the Jinja (SLS) context .

AFAIU this PR is tackling problem `1` but not problem `2`.

I see your point that we don't really want to continue growing hardcoded Zypper versions checks on every new Zypper feature we want to implement in Salt. This would need a better thinking and refactor.

About problem `2`, this is something to keep thinking, but I think we shouldn't make this PR to require it. I forgot about https://github.com/saltstack/salt/pull/48152 :sweat_smile: 

Since we have been dealing with this PR since a while already, and it's currently blocking other implementations. I would say that we could accept this current implementation now, and continue with the refactor in a following-up PR.

salt/modules/zypperpkg.py
brejoc: Good point. We'll address this in a follow-up PR. Thx!

brejoc (approved):
In general this gets a :+1: from me, but there will be additional changes coming later. But not in this PR.

salt/modules/zypperpkg.py
isbm: @meaksh yes, exactly. The problem `#1` is just "remove these if/else, but add two lines and you're done for now". But *also* I pointed out so for the *future* we need to keep in mind the problem `#2`, as you correctly pointed out.

PR #330, opened 2021-03-11
Update target fix for salt-ssh and avoiding race condition on salt-ssh event processing (bsc#1179831, bsc#1182281) - 3000

vzhestkov:
### What does this PR do?

[1179831](https://github.com/SUSE/spacewalk/issues/14002) - L3: Scheduled SSH-Push Tasks throws Exception and cause Salt-API to fail
[1182281](https://github.com/SUSE/spacewalk/issues/14002) - L3: SSH-Push/Salt-SSH "Adding minions for job xyz: [None]" cause unhandled Exception and salt-api to crash

### Previous Behavior
Traceback caused by improper processing of `tgt` parameter in case if it's a list:
```
...
File "/usr/lib/python3.6/site-packages/salt/client/ssh/__init__.py", line 233, in __init__
self._update_targets()
File "/usr/lib/python3.6/site-packages/salt/client/ssh/__init__.py", line 411, in _update_targets
if salt.utils.network.is_reachable_host(hostname):
...
```

And the traceback caused by race condition on deletion `roster_file` option during event processing for salt-ssh with large number of threads:
```
...
File "/usr/lib/python3.6/site-packages/salt/roster/__init__.py", line 40, in get_roster_file
del options['roster_file']
KeyError: 'roster_file'
...
```

### New Behavior
Normal behavior with no traceback. Additional warning message was implemented to show salt-ssh which can't be resolved.

### Tests written?
No

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -370,7 +370,7 @@ class SSH(object):
                             self.__parsed_rosters[self.ROSTER_UPDATE_FLAG] = False
                             return
 
-    def _update_roster(self):
+    def _update_roster(self, hostname=None, user=None):
         '''
         Update default flat roster with the passed in information.
         :return:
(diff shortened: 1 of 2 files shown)


agraul:
@vzhestkov is 387c508 part of the PTFs or was that done after the fact?

vzhestkov:
It was not a part of PTF, I made this change after investigating the next issue this customer found and do some checks on my local test system to improve the fix in this part. One of the issue the customer faced sometimes is still there, but we are looking for the exact conditions causing it, but there is other BSC for it.

We can check new PTF made of these 3 commits, but in general it should work almost the same way as the original PTF.

PR #336, opened 2021-03-16
Update target fix for salt-ssh to process targets list (bsc#1179831)

vzhestkov:
### What does this PR do?

[1179831](https://github.com/SUSE/spacewalk/issues/14002) - L3: Scheduled SSH-Push Tasks throws Exception and cause Salt-API to fail
[1182281](https://github.com/SUSE/spacewalk/issues/14002) - L3: SSH-Push/Salt-SSH "Adding minions for job xyz: [None]" cause unhandled Exception and salt-api to crash - FIXED IN UPSTREAM (NO CHANGES in PR for 3002.2)

### Previous Behavior
Traceback caused by improper processing of `tgt` parameter in case if it's a list:
```
...
File "/usr/lib/python3.6/site-packages/salt/client/ssh/__init__.py", line 233, in __init__
self._update_targets()
File "/usr/lib/python3.6/site-packages/salt/client/ssh/__init__.py", line 411, in _update_targets
if salt.utils.network.is_reachable_host(hostname):
...
```

### New Behavior
Normal behavior with no traceback. Additional warning message was implemented to show salt-ssh which can't be resolved.

### Tests written?
No

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -376,7 +376,7 @@ class SSH:
                             self.__parsed_rosters[self.ROSTER_UPDATE_FLAG] = False
                             return
 
-    def _update_roster(self):
+    def _update_roster(self, hostname=None, user=None):
         """
         Update default flat roster with the passed in information.
         :return:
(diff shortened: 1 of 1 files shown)


moio:
@vzhestkov was this case a regression - ie. it started malfunctioning when we pushed out a new Salt version which was broken upstream?

vzhestkov:
No, the issue was there in 3000 and 3002.2. The fix is almost the same for 3000 and 3002.2, but 3002.2 has no issue causing exception due to race condition and this fix was backported to 3000.

PR #337, opened 2021-03-18
[WIP] Notify beacon for Debian/Ubuntu systems

ricardoasmarques:
### What does this PR do?

This PR implements an `aptnotify` that will trigger a "package refresh" on Uyuni when `apt-get` is used to manually install/update/remove packages (similar to the existing `zyppnotify` and `yumnotify`).

**TODO**

- [ ] Packaging: The `99aptnotify` file must be installed in `/etc/apt/apt.conf.d/99aptnotify`
- [ ] Packaging: The `aptnotify` file must be installed in `/usr/bin/aptnotify`

- https://github.com/SUSE/spacewalk/issues/11561

### Previous Behavior

### New Behavior

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/scripts/suse/apt/aptnotify
+++ b/scripts/suse/apt/aptnotify
@@ -1,0 +1,46 @@
+#!/usr/bin/python3
+
+import os
+import hashlib
+
+CK_PATH = "/var/cache/salt/minion/rpmdb.cookie"
+DPKG_PATH = "/var/lib/dpkg/status"
+
+def _get_mtime():
+    """
+    Get the modified time of the Package Database.
+
+    Returns:
+        Unix ticks
+    """
+    return os.path.exists(DPKG_PATH) and int(os.path.getmtime(DPKG_PATH)) or 0
+
+
+def _get_checksum():
+    """
+    Get the checksum of the Package Database.
+
+    Returns:
+        hexdigest
+    """
+    digest = hashlib.sha256()
+    with open(DPKG_PATH, "rb") as rpm_db_fh:
+        while True:
+            buff = rpm_db_fh.read(0x1000)
+            if not buff:
+                break
+            digest.update(buff)
+    return digest.hexdigest()
+
+
+def dpkg_post_invoke():
+    """
+    Hook after the package installation transaction.
+    """
+    if 'SALT_RUNNING' not in os.environ:
+        with open(CK_PATH, 'w') as ck_fh:
+            ck_fh.write('{chksum} {mtime}\n'.format(chksum=_get_checksum(), mtime=_get_mtime()))
+
+
+if __name__ == "__main__":
+    dpkg_post_invoke()
--- a/scripts/suse/apt/99aptnotify
+++ b/scripts/suse/apt/99aptnotify
@@ -1,0 +1,1 @@
+DPkg::Post-Invoke {"/usr/bin/aptnotify";};


meaksh:
NOTE: As mentioned in the **TODO**, this new file needs to be deployed when installing the DEB packages. This requires we change the `debian.tar` in our OBS debian subprojects to make this file to be included.

We will be doing it ourselves at the time of merging this PR.

scripts/suse/apt/aptnotify
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python3
```
meaksh, line 1: Just wondering: is this python3 shebang causing any pain to a Python 2, in for example Ubuntu 16.04?

scripts/suse/apt/aptnotify
```diff
@@ -0,0 +1,46 @@
...
+
+def _get_checksum():
+    """
+    Get the checksum of the Package Database.
+
+    Returns:
+        hexdigest
+    """
+    digest = hashlib.sha256()
+    with open(DPKG_PATH, "rb") as rpm_db_fh:
```
meaksh, line 27: Minor nitpick:

```suggestion
    with open(DPKG_PATH, "rb") as pkg_db_fh:
```

scripts/suse/apt/aptnotify
```diff
@@ -0,0 +1,46 @@
...
+        while True:
+            buff = rpm_db_fh.read(0x1000)
```
meaksh, line 29:
```suggestion
            buff = pkg_db_fh.read(0x1000)
```

scripts/suse/apt/aptnotify
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python3
+
+import os
+import hashlib
+
+CK_PATH = "/var/cache/salt/minion/rpmdb.cookie"
```
agraul, line 6: I think this should be 
```suggestion
CK_PATH = "/var/cache/salt/minion/dpkg.cookie"
```

scripts/suse/apt/aptnotify
```diff
@@ -0,0 +1,46 @@
...
+                break
+            digest.update(buff)
+    return digest.hexdigest()
+
+
+def dpkg_post_invoke():
+    """
+    Hook after the package installation transaction.
+    """
+    if 'SALT_RUNNING' not in os.environ:
```
agraul, line 40: We need to make sure that `SALT_RUNNING` is set for this to work

meaksh: Hmmm, not in this case. Here we're in the APT plugin, which always run when there is apt transaction.

In this case, we want to avoid changing the cookie if the `SALT_RUNNING` variable is set, and it's only been set by Salt aptpkg module when calling apt.

Changing the cookie will trigger the beacon event and therefore a pkg list refresh, and we don't want that when the apt execution was made by Salt.

scripts/suse/apt/aptnotify
meaksh: You're right, this is a good point.

I think they decided to not changing this filename at this point because that would require refactors on the Java backend that generated the pkgset beacon configuration.

Currently, the configuration for the pkgset beacon is always defining `/var/cache/salt/minion/rpmdb.cookie` as the source for the cookie, regardless the OS.

scripts/suse/apt/aptnotify
agraul: My comment was too easy to misunderstand. Let me try again ;)

We need to make sure that aptpkg is setting `SALT_RUNNING`, just like yumpkg and zypperpkg do. aptpkg uses the same mechanism that zypperpkg does, i.e. setting `SALT_RUNNING` can be done in the configuration file. I didn't check if it is already configured to do so, but I think not (given the bug report we had because we didn't configure it for zypperpkg). Either way, we should not forget this step.

scripts/suse/apt/aptnotify
agraul: What about a generic name in that case? Could be named after the virtual pkg module: "pkg.cookie". Looking at the [Java code](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java#L47), the name should be easy to change.

rjmateus: Hey, I agree we should have a generic name but I'm not sure what will happen after we change it.
The problem would be for existing costumer with everything already in place. We would force them to upgrade all the clients tools to generate the cookie file with the new name or beacon will not work after change the java code side.
Do we really want to figure out what will be all the side effects of changing it?

isbm: Nothing special will happen if you change it. Just make sure you update minion *and* the backend.

mcalmer: But we cannot guarantee that the custom update both at the same time. I am not sure what we do exactly on the java side, but maybe we can do it for both names, old and new. After some time we can remove the old name and only leave the generic name.

paususe: Stupid proposal: symlink :-)

hey @isbm , glad to see you around!

paususe:
I know it's my fault but shouldn't this be called `dpkgnotify` instead of `aptnotify`? After all, it should run after dpkg, since you may install/uninstall/update a package manually with dpkg (same as you can with rpm).

meaksh:
I think this can be closed in favor of https://github.com/openSUSE/salt/pull/347 (which is already approved). I'm closing this draft PR now.

Feel free to reopen if necessary. Thanks!

PR #347, opened 2021-03-30
Notify beacon for Debian/Ubuntu systems

rjmateus:
### What does this PR do?

This PR implements an dpkgnotify that will trigger a "package refresh" on Uyuni when apt-get is used to manually install/update/remove packages (similar to the existing zyppnotify and yumnotify).
It's a replacement of https://github.com/openSUSE/salt/pull/337 with the comments addressed.

Fixes: https://github.com/SUSE/spacewalk/issues/11561

### Previous Behavior

### New Behavior

### Tests written?
No

### Commits signed with GPG?

Yes


scripts/suse/dpk/README.md
```diff
@@ -0,0 +1,9 @@
+## What it is
+
+Debian base package to notify instalation of new packages outside the control of uyuni/suse manager
```
agraul, line 3: Strictly speaking the script will collect dpkg operation outside of Salt, not Uyuni or SUSE Manager. 
```suggestion
Debian base package to notify installation of new packages outside the control of Uyuni/SUSE Manager
```

agraul:
I would move/rename the directory from `dpk` to `dpkg`.

rjmateus:
Thanks for the review @agraul 
I have correct the suggestions

agraul (approved):

agraul:
One more question @rjmateus: Is this not working with Python2 at all or was it just not tested?

rjmateus:
It will not work with python2, since the script is including `#!/usr/bin/python3`. If you think this can be a problem, then we need to change the script. Thanks for the review

agraul:
Right, with the current shebang this won't run on Python 2. I wondered if this was on purpose to exclude Python 2 or if it was done because Python 3 is the default these days. I don't think we can change it in this here, it needs to be done at build-time. I'm looking into how to do it, our debian packages are currently built without patching.

PR #355, opened 2021-04-20
Move vendor change logic to zypper class

mseidl:
### What does this PR do?
Moves vendor change logic to the zypper class and re-organizes some tests, as well as adds the allowvendorchange option.  

We need to fix one of the tests.  

https://github.com/SUSE/spacewalk/issues/14313
Upstream PR: https://github.com/saltstack/salt/pull/58560

### Previous Behavior

### New Behavior

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1463,7 +1463,8 @@ def install(name=None,
         DEPRECATED(use allowvendorchange): If set to True, do not allow vendor changes. Default: True
 
     allowvendorchange
-        If set to True, allow vendor changes. Default: False
+        If set to True, vendor change is allowed. Default: False
+        If both allowvendorchange and novendorchange are passed, only allowvendorchange is used.
 
     version
         Can be either a version number, or the combination of a comparison
@@ -1713,6 +1714,7 @@ def upgrade(refresh=True,
 
     allowvendorchange
         If set to True, vendor change is allowed. Default: False
+        If both allowvendorchange and novendorchange are passed, only allowvendorchange is used.
 
     skip_verify
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -355,10 +355,15 @@ class _Zypper(object):
             if self.__systemd_scope:
                 cmd.extend(['systemd-run', '--scope'])
             cmd.extend(self.__cmd)
+
             if self.avc:
                 for i in ["install", "upgrade", "dist-upgrade"]:
                     if i in cmd:
-                        cmd.insert(cmd(index) + 1, "--allow-vendor-change")
+                        if i == "install" and self.inst_avc:
+                           cmd.insert(cmd.index(i) + 1, "--allow-vendor-change")
+                        elif i in ["upgrade", "dist-upgrade"] and self.dup_avc:
+                           cmd.insert(cmd.index(i) + 1, "--allow-vendor-change")
+
             log.debug("Calling Zypper: %s", ' '.join(cmd))
             self.__call_result = __salt__['cmd.run_all'](cmd, **kwargs)
             if self._check_result():
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -355,8 +355,6 @@ class _Zypper(object):
             if self.__systemd_scope:
                 cmd.extend(['systemd-run', '--scope'])
             cmd.extend(self.__cmd)
-            log.error("j**************************8")
-            log.error(cmd)
             if self.avc:
                 for i in ["install", "upgrade", "dist-upgrade"]:
                     if i in cmd:
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1460,7 +1460,7 @@ def install(name=None,
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
 
     novendorchange
-        Disallow vendor change
+        DEPRECATED(use allowvendorchange): If set to True, do not allow vendor changes. Default: True
 
     allowvendorchange
         If set to True, allow vendor changes. Default: False
@@ -1709,10 +1709,10 @@ def upgrade(refresh=True,
         Specify a list of package repositories to upgrade from. Default: None
 
     novendorchange
-        If set to True, no allow vendor changes. Default: False
+        DEPRECATED(use allowvendorchange): If set to True, do not allow vendor changes. Default: True
 
     allowvendorchange
-        If set to True, allow vendor changes. Default: False
+        If set to True, vendor change is allowed. Default: False
 
     skip_verify
         Skip the GPG verification check (e.g., ``--no-gpg-checks``)
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -105,10 +105,6 @@ class _Zypper(object):
     ZYPPER_LOCK = '/var/run/zypp.pid'
     TAG_RELEASED = 'zypper/released'
     TAG_BLOCKED = 'zypper/blocked'
-    # Dist upgrade vendor change support (SLE12+)
-    dup_avc = False
-    # Install/Patch/Upgrade vendor change support (SLE15+)
-    inst_avc = False
 
     def __init__(self):
         '''
@@ -137,6 +133,13 @@ class _Zypper(object):
         self.__ignore_not_found = False
         self.__systemd_scope = False
         self.__root = None
+
+        # Dist upgrade vendor change support (SLE12+)
+        self.dup_avc = False
+        # Install/Patch/Upgrade vendor change support (SLE15+)
+        self.inst_avc = False
+        # Flag if allow vendor change should be allowed
+        self.avc = False
 
         # Call status
         self.__called = False
@@ -182,6 +185,8 @@ class _Zypper(object):
             self.__no_raise = True
         elif item == 'refreshable':
             self.__refresh = True
+        elif item == 'allow_vendor_change':
+            return self.__allow_vendor_change
         elif item == 'call':
             return self.__call
         else:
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_zypperpkg.py
```diff
@@ -564,17 +569,45 @@ def test_upgrade_success(self):
...
-                            "dist-upgrade",
-                            "--auto-agree-with-licenses",
-                            "--dry-run",
-                            "--from",
-                            "Dummy",
-                            "--from",
-                            "Dummy2",
-                            "--debug-solver",
-                        )
+
```
brejoc, line 572:
```suggestion
                        )
```

A closing bracket is missing here.

brejoc (changes requested):

tests/unit/modules/test_zypperpkg.py
```diff
@@ -450,6 +452,115 @@ def test_latest_version(self):
...
+                            ret = zypper.upgrade(
+                                dist_upgrade=True,
+                                dryrun=True,
+                                fromrepo=["Dummy", "Dummy2"],
+                                novendorchange=False,
+                            )
+                            # Since the "call" method is completely mocked away, we need to
+                            # assert on the vendor_change_mock, because the logic of the vendor-change
+                            # flag insertion is in thes __call method.
+                            vendor_change_mock.assert_any_call(False, False, True)
```
brejoc, line 513: Maybe we should also assert `zypper.__zypper__.avc` for those. :thinking:

tests/unit/modules/test_zypperpkg.py
```diff
@@ -450,6 +452,117 @@ def test_latest_version(self):
...
+                                        'lowpkg.version_cmp': MagicMock(return_value=1)}):
+                            ret = zypper.upgrade(
+                                dist_upgrade=True,
+                                dryrun=True,
+                                fromrepo=["Dummy", "Dummy2"],
+                                novendorchange=False,
+                            )
+                            # Since the "call" method is completely mocked away, we need to
+                            # assert on the vendor_change_mock, because the logic of the vendor-change
+                            # flag insertion is in thes __call method.
```
agraul, line 512: `call` is replaced by `zypper_mock`, I think we should check that it is called correctly. `vendor_change_mock` is an implementation detail, I don't think checking is it important. 

I think the goal of this test is to verify that a `zypper.upgrade(xzy, nodeverchange=False)` results in a specific `zypper dist-upgrade` call, isn't it?

tests/unit/modules/test_zypperpkg.py
```diff
@@ -450,6 +452,117 @@ def test_latest_version(self):
...
+
+    def test_upgrade_with_allowvendorchange(self):
+        '''
+        Perform dist-upgrade with allowvendorchange set to True.
+        '''
+        with patch.dict(zypper.__grains__, {'osrelease_info': [12, 1]}), \
+                patch('salt.modules.zypperpkg.refresh_db', MagicMock(return_value=True)), \
+                patch('salt.modules.zypperpkg._systemd_scope', MagicMock(return_value=False)):
+            with patch('salt.modules.zypperpkg.__zypper__.noraise.call', MagicMock()) as zypper_mock:
+                with patch('salt.modules.zypperpkg.__zypper__.allow_vendor_change', MagicMock()) as vendor_change_mock:
```
agraul, line 525: Same as above, I think the test loses a lot of meaning if the vendor change logic is replaced by a mock

tests/unit/modules/test_zypperpkg.py
```diff
@@ -450,6 +452,117 @@ def test_latest_version(self):
...
+                                allowvendorchange=True,
+                            )
+                            # Since the "call" method is completely mocked away, we need to
+                            # assert on the vendor_change_mock, because the logic of the vendor-change
+                            # flag insertion is in thes __call method.
+                            assert zypper.__zypper__.avc == True
+                            vendor_change_mock.assert_any_call(True, True, True)
+
+    def test_upgrade_old_zypper(self):
+        with patch.dict(zypper.__grains__, {'osrelease_info': [12, 1]}), \
```
agraul, line 547: This uses the same `osrelease_info` as the other upgrade tests. Is that correct?

agraul (changes requested):
I think it would be good to document how `allowvendorchange` and `novendorchange` play together. I also left some remarks inline, both for the implementation and the tests.

salt/modules/zypperpkg.py
```diff
@@ -222,6 +227,16 @@ def stderr(self):
     def pid(self):
         return self.__call_result.get('pid', '')
 
+    def __allow_vendor_change(self, allowvendorchange, novendorchange, dist_upgrade=False):
+        if allowvendorchange or not novendorchange:
+            self.refresh_zypper_flags()
+            if self.dup_avc or self.inst_avc:
```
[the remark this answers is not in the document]
mseidl: We have 2 flags: for dist upgrade which works with sle12 and instlall/patch/upgrade which only works with sle15+

salt/modules/zypperpkg.py
```diff
@@ -222,6 +227,16 @@ def stderr(self):
     def pid(self):
         return self.__call_result.get('pid', '')
 
+    def __allow_vendor_change(self, allowvendorchange, novendorchange, dist_upgrade=False):
```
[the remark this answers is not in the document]
mseidl: The issue is, The novendorchange option doesn't make sense as by default zypper doesn't allow vendor change.  The original zypperpkg would add a --no-allow-vendor-change or nothing which would never enable vendor change. The novendorchange should be deprecated.    and novendorchange=False for vendor change just gets weird.

salt/modules/zypperpkg.py
```diff
@@ -1689,6 +1706,9 @@ def upgrade(refresh=True,
     novendorchange
         If set to True, no allow vendor changes. Default: False
 
+    allowvendorchange
+        If set to True, allow vendor changes. Default: False
+
```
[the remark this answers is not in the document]
isbm: Actually this is superconfusing, if you would ask me. :wink: At the end, the vendor is changed or not, regardless what version of SLE. I would just `vendor_change:bool = False` and grains your friend to figure out how to apply to a specific SLE version.

Just my 1 cent here.

salt/modules/zypperpkg.py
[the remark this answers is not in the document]
isbm: In these kind of situations I would keep looking from the end user perspective: what would you seek to achieve at the end, as user? That's the point of wrapping all that stuff with a Salt module and bring a simple and common interface of achieving desired result on different SLE versions, those are so much different inside. That said, given function is just that: to allow vendor change or not. Therefore in YAML I would ask "Would you like to change the vendor?".

```yaml
...
   change_vendor: true
```

...good! So I will take that information and would then use grains and figure out these combination, at the end I would say "The vendor has been changed".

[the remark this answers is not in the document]
mseidl: removed

tests/unit/modules/test_zypperpkg.py
mseidl: We fixed it. That was the old check which was just the os release.

tests/unit/modules/test_zypperpkg.py
```diff
@@ -470,16 +470,21 @@ def test_upgrade_without_vendor_change(self):
...
         '''
-        with patch.dict(zypper.__grains__, {'osrelease_info': [12, 1]}), \
-                patch('salt.modules.zypperpkg.refresh_db', MagicMock(return_value=True)), \
+        with patch('salt.modules.zypperpkg.refresh_db', MagicMock(return_value=True)), \
+                patch('salt.modules.zypperpkg.__zypper__.refresh_zypper_flags', MagicMock()) as refresh_flags_mock, \
                 patch('salt.modules.zypperpkg._systemd_scope', MagicMock(return_value=False)):
             with patch('salt.modules.zypperpkg.__zypper__.noraise.call', MagicMock()) as zypper_mock:
                 with patch('salt.modules.zypperpkg.list_pkgs', MagicMock(side_effect=[{"vim": "1.1"}, {"vim": "1.2"}])):
                     ret = zypper.upgrade(dist_upgrade=True, novendorchange=True)
+                    refresh_flags_mock.assert_not_called()
```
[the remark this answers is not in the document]
agraul: On a second though, maybe this assertion is not needed at all. The assertion below is already checking that no vendor change flag is passed to Zypper

tests/unit/modules/test_zypperpkg.py
agraul: The `vendor_change_mock` does not exist anymore, rendering this comment outdated.

tests/unit/modules/test_zypperpkg.py
```diff
@@ -541,11 +710,41 @@ def test_upgrade_with_allowvendorchange(self):
                             # assert on the vendor_change_mock, because the logic of the vendor-change
```
agraul, line 710: Same as above, there is no `vendor_change_mock`

tests/unit/modules/test_zypperpkg.py
```diff
@@ -541,11 +710,41 @@ def test_upgrade_with_allowvendorchange(self):
...
+                            zypper.__zypper__.inst_avc = True
+                            zypper.__zypper__.dup_avc = True
+                            ret = zypper.upgrade(
+                                dist_upgrade=True,
+                                dryrun=True,
+                                fromrepo=["Dummy", "Dummy2"],
+                                allowvendorchange=False,
+                            )
+                            # Since the "call" method is completely mocked away, we need to
+                            # assert on the vendor_change_mock, because the logic of the vendor-change
```
agraul, line 740: Same as above, there is no `vendor_change_mock`

agraul (changes requested):
I still have doubts about the design of the two vendor-change related flags. To me it is not an easy to use abstraction over `zypper`'s CLI.

Overall the tests seem better, but the readability of the tests could be improve a bit using e.g. pytest fixtures and parameterization. I also wonder if we need all of those new test, e.g. both `test_zypper_call_dist_upgrade_with_avc_true` and `test_upgrade_with_allowvendorchange_true`. Are those really testing different things or just different levels of the same thing ? Is the lower-level one needed or already covered by the higher level one?

agraul (approved):

PR #365, opened 2021-05-06
Check if dpkgnotify is executable (bsc#1186674) - 3000

vzhestkov:
### What does this PR do?

In case of removing `salt-minion` package from the system `apt` returns the error messages that it's impossible to execute the `dpkgnotify` as the apt config to invoke this call is still active.

### Previous Behavior
`apt` errors on removing `salt-minion` from the system

### New Behavior
No errors

### Tests written?
No

### DIFF
--- a/scripts/suse/dpkg/99dpkgnotify
+++ b/scripts/suse/dpkg/99dpkgnotify
@@ -1,1 +1,1 @@
-DPkg::Post-Invoke {"/usr/bin/dpkgnotify";};
+DPkg::Post-Invoke {"if [ -x /usr/bin/dpkgnotify ]; then /usr/bin/dpkgnotify; fi;";};


agraul:
I think this does not hurt, but since we control the deployment of `/usr/bin/dpkgnotify` via our packages, it shouldn't be needed. We need to ensure that `/usr/bin/dpkgnotify` is executable.

vzhestkov:
The problem that on removing `salt-minion` `apt` is invoking the specified section after removing the package and `/usr/bin/dpkgnotify` is missing in time when `apt` trying to invoke it as the config file was loaded by `apt` before removing the package. It sould be fixed even if we are removing this config file.

agraul:
Ok, I wasn't aware of that race condition. Makes sense! Could you add that to the commit message please?

vzhestkov:
@agraul, Is it better now?

agraul:
I was more thinking along the lines of having a message body, e.g.
```
Only invoke dpkgnotify if it is executable

There is a race condition on the removal of the salt-minion package.
The salt-minion package includes the /usr/bin/dpkgnotify script, so when
it is uninstalled the script is removed. apt still tries to call the
script after removing it, causing an error message.

This change ensures that apt only calls the script if it exists and is
executable to avoid the race condition.
```

That way it is easy to understand when you go through the git log in 6 months from now because you have all the context you need. The above is only an example of how I would phrase it.

agraul (approved):

PR #373, opened 2021-05-20
Handle volumes on stopped pools in virt.vm_info

cbosdo:
Backport of https://github.com/saltstack/salt/pull/60133

For VMs having at least a disk on a stopped volume, we don't want the
user to get an exception when running virt.vm_info. Instead just provide
less information.

### DIFF
--- a/changelog/60132.fixed
+++ b/changelog/60132.fixed
@@ -1,0 +1,1 @@
+Gracefuly handle errors in virt.vm_info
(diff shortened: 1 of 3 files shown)


cbosdo:
https://bugzilla.suse.com/show_bug.cgi?id=1186287

meaksh (approved):

PR #379, opened 2021-06-10
Add venv executor

vzhestkov:
### What does this PR do?

This executor prevents passing `PYTHONHOME` environment variable
to child processes from venvjailed salt-minion

The change is only required for `3002.2` as a base of the salt-minion bundle.

Requires additional config section for salt minion config:
```
module_executors:
  - venv
```

### Previous Behavior
Possible tracebacks on calling external python scripts from `salt-minion`.
For examle: `dnf` called with `yumpkg` or `dpkgnotify` with `aptpkg` and any other python script could fail on calling with `cmd` salt module.

### New Behavior

### Tests written?
Yes

### DIFF
--- a/salt/executors/venv.py
+++ b/salt/executors/venv.py
@@ -1,0 +1,13 @@
+"""
+Wipe PYTHONHOME environment variable for venvjailed salt-minion
+"""
+import os
+
+
+def execute(opts, data, func, args, kwargs):
+    """
+    Wipe PYTHONHOME environment variable for venvjailed salt-minion
+    """
+    if os.environ.get("VIRTUAL_ENV", None) == os.environ.get("PYTHONHOME", ""):
+        os.environ.pop("PYTHONHOME", None)
+    return __executors__["direct_call.execute"](opts, data, func, args, kwargs)
--- a/tests/pytests/unit/executors/test_venv.py
+++ b/tests/pytests/unit/executors/test_venv.py
@@ -1,0 +1,43 @@
+"""
+Unit test for venv executor
+"""
+
+import pytest
+
+import os
+
+import salt.executors.venv as venv_exec
+
+
+@pytest.fixture(scope="session", autouse=True)
+def executors():
+    env_save = dict(os.environ)
+    venv_exec.__executors__ = {"direct_call.execute": direct_call_helper}
+    yield
+    os.environ.clear()
+    os.environ.update(env_save)
+
+
+def direct_call_helper(_opts, _data, _func, _args, _kwargs):
+    return
+
+
+def test_wipe_PYTHONHOME_environment_variable_for_venvjailed_minion(executors):
+    """
+    In case of using salt-minion in the venvjailed environment
+    PYTHONHOME environment variable should be wiped
+    """
+    os.environ["PYTHONHOME"] = os.environ["VIRTUAL_ENV"] = "/opt/venv-salt-minion"
+    venv_exec.execute(None, None, None, None, None)
+    assert "PYTHONHOME" not in os.environ
+
+
+def test_no_wipe_PYTHONHOME_environment_variable_for_non_venvjailed_minion(executors):
+    """
+    In case of using salt-minion in the non venvjailed environment
+    PYTHONHOME environment variable shouldn't be changed
+    """
+    test_path = "/usr"
+    os.environ["PYTHONHOME"] = test_path
+    venv_exec.execute(None, None, None, None, None)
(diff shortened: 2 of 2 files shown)


agraul:
What exactly do the tracebacks look like? I don't fully understand what exactly is broken. Could you please paste an example or two?

The below is based on my incomplete understanding of the problem, but I still want to mention two things


1) How would this work with `pip`. Salt upstream supports `pip install` in their tiamat binaries (at least that's the plan, I haven't checked its current status) and I think we should think about that as well. It would allow bundle users running `venv-salt-minion` to use arbitrary Salt modules, not just the ones that already have their dependencies in the bundle.

2) Would it be worth it to have some way to configure this executor? I am thinking about something similar to the transactional-update executor, maybe in a stripped-down, simpler way (e.g. a set of execution/state modules that are included or excluded). Would that make sense or would it be pointless?

salt/executors/venv.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
agraul, line 1: This shouldn't be needed for Python 3

tests/pytests/unit/executors/test_venv.py
```diff
@@ -0,0 +1,43 @@
...
+    assert "PYTHONHOME" not in os.environ
+
+
+def test_no_wipe_PYTHONHOME_environment_variable_for_non_venvjailed_minion(executors):
+    """
+    In case of using salt-minion in the non venvjailed environment
+    PYTHONHOME environment variable shouldn't be changed
+    """
+    test_path = "/usr"
+    os.environ["PYTHONHOME"] = test_path
```
agraul, line 41: Could this interfere with other unit tests?

vzhestkov: No, there are workaround for it in the `fixture`

salt/executors/venv.py
vzhestkov: Thanks. Removed

salt/executors/venv.py
```diff
@@ -0,0 +1,13 @@
...
+Wipe PYTHONHOME environment variable for venvjailed salt-minion
+"""
+import os
+
+
+def execute(opts, data, func, args, kwargs):
+    """
+    Wipe PYTHONHOME environment variable for venvjailed salt-minion
+    """
+    if os.environ.get("VIRTUAL_ENV", None) == os.environ.get("PYTHONHOME", ""):
```
aplanas, line 11: I am confused with both have different defaults (one `None` and the other `""`). I would be super super simple here:

* Do not upstream the executor in salt, is a very specific use case and fairly useless outside the venv-salt-minion
* Add the executor only inside the venv-salt-minion package, and add the configuration snipped that activate it inside the venv
* Make this executor a two line code:

```python
# No exception if it is not there
os.environ.pop("PYTHONHOME", None)
return __executors__["direct_call.execute"](opts, data, func, args, kwargs)
```
* Drop the tests

aplanas (changes requested):

salt/executors/venv.py
vzhestkov: @aplanas, yes you are right it's useless for any case other than `venv-salt-minion`, but I suppose it's better to have it here (not in the upstream salt for sure) just to make it easier to maintain.
`os.environ.get("VIRTUAL_ENV", None) == os.environ.get("PYTHONHOME", "")` was an attempt to prevent the conditional step if both are not present. It's safe for sure, but here I'm trying to prevent wiping `PYTHONHOME` if the executor is used outside the `venv-salt-minion`. I can't imagine such case anyway and probably we don't really need this check.

aplanas: > but I suppose it's better to have it here (not in the upstream salt for sure) just to make it easier to maintain

Uhm I would not put in the salt package, as the code would be distributed for all the users and will require an update of the salt package each time that we want to fix something in the venv-salt-minion package.

Also making the two-line version a thing inside the venv-salt-minion open the door to provide more similar fixes for this specific use case without any change. For example if we want to adjust the "pip" command we can influence from here.

> `os.environ.get("VIRTUAL_ENV", None) == os.environ.get("PYTHONHOME", "")` was an attempt to prevent the conditional step if both are not present.

Sure. I was wondering about the different defaults (like making both `None`, or both `""`). And this is not needed if lives in the venv (as we want to be removed always)

vzhestkov: Agree, closing the PR without merging

PR #386, opened 2021-06-24
Enhance openscap module: add "xccdf_eval" call

meaksh:
### What does this PR do?

This PR introduces a new function for the openscap execution module: `xccdf_eval`

This function, allows to pass more parameters than the `xccdf` function, like `--fetch-remote-resources`, `--rule`, `--remediate`, `--tailoring-file`, `--tailoring-id`, etc.

This PR keeps the old `xccdf` function so this shouldn't break or harm anyone that is using the old function of the module.

**IMPORTANT:** There is an upstream PR here https://github.com/saltstack/salt/pull/60211 that still needs some more work in order to not remove the old function and properly deprecate it. Once this upstream PR is fixed we would probably want to switch to the final `xccdf` function.

### Tests written?
Please review the [test documentation](https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html) for details on how to implement tests into Salt's test suite.

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/changelog/59756.added
+++ b/changelog/59756.added
@@ -1,0 +1,1 @@
+adding new call for openscap xccdf eval supporting new parameters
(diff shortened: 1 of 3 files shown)


vzhestkov (approved):
Looks good

PR #387, opened 2021-06-25
Implementation of held/unheld functions for state pkg

vzhestkov:
### What does this PR do?

Port of https://github.com/saltstack/salt/pull/56935 and https://github.com/saltstack/salt/pull/60432
Fixes https://github.com/SUSE/spacewalk/issues/15258 (bsc#1187813)

Implements `held`/`unheld` functions for state `pkg` module.
Examples:
```
sysstat:
  pkg.held: []
```
```
held-pkgs:
  pkg.held:
    - pkgs:
      - sysstat
      - vnstat: 1.17-bp153.1.18
```
```
held-pkgs-clear:
  pkg.held:
    - pkgs: []
    - replace: True
```
```
unheld-pkgs:
  pkg.unheld:
    - pkgs:
      - sysstat
      - vnstat: 1.17-bp153.1.18
      - test-package
```
```
unheld-all:
  pkg.unheld:
    - all: True
```

Implements `list_holds` function for `zypperpkg` module. The output is similar to `yumpkg.list_holds`. The only difference is that `zypperpkg` calculates the current locked packages from `list_locks` as the yum and zypper have different way of managing locks.

Improves `hold` function for `zypperpkg` to handle package version if specified.

### Previous Behavior
Tracebacks on calling `zypperpkg.hold`/`zypperpkg.unhold`

### New Behavior
Expected behavior and implementation of `pkg.held` and `pkg.unheld`.

### Tests written?
No

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -2071,6 +2071,76 @@ def purge(
     return _uninstall(inclusion_detection, name=name, pkgs=pkgs, root=root)
 
 
+def list_holds(pattern=None, full=True, root=None, **kwargs):
+    """
+    List information on locked packages.
+
+    .. note::
+        This function returns the computed output of ``list_locks``
+        to show exact locked packages.
+
+    pattern
+        Regular expression used to match the package name
+
+    full : True
+        Show the full hold definition including version and epoch. Set to
+        ``False`` to return just the name of the package(s) being held.
+
+    root
+        Operate on a different root directory.
+
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' pkg.list_holds
+        salt '*' pkg.list_holds full=False
+    """
+    locks = list_locks(root=root)
+    ret = []
+    inst_pkgs = {}
+    for solv_name, lock in locks.items():
+        if lock.get("type", "package") != "package":
+            continue
+        try:
+            found_pkgs = search(
+                solv_name,
+                root=root,
+                match=None if "*" in solv_name else "exact",
+                case_sensitive=(lock.get("case_sensitive", "on") == "on"),
+                installed_only=True,
+                details=True,
+                all_versions=True,
+                ignore_no_matching_item=True,
+            )
+        except CommandExecutionError:
+            continue
+        if found_pkgs:
+            for pkg in found_pkgs:
+                if pkg not in inst_pkgs:
+                    inst_pkgs.update(
+                        info_installed(
+                            pkg, root=root, attr="edition,epoch", all_versions=True
+                        )
+                    )
+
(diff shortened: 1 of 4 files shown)


salt/modules/zypperpkg.py
```diff
@@ -2143,44 +2211,67 @@ def clean_locks(root=None):
...
+                ret[target]["changes"]["old"] = "hold"
+                ret[target]["comment"] = "Package {} is no longer held.".format(target)
         else:
-            missing.append(pkg)
-            ret[pkg]["comment"] = "Package {} unable to be unheld.".format(pkg)
+            ret[target]["comment"] = "Package {} was already unheld.".format(target)
 
     if removed:
-        __zypper__(root=root).call("rl", *removed)
+        __zypper__.call("rl", *removed)
```
meaksh, line 2274: Shouldn't we keep `(root=root)` here?

salt/modules/zypperpkg.py
```diff
@@ -2225,48 +2316,55 @@ def remove_lock(packages, root=None, **kwargs):  # pylint: disable=unused-argume
...
+            ret[target]["comment"] = "Package {} is now being held.".format(target)
         else:
-            ret[pkg]["comment"] = "Package {} is already set to be held.".format(pkg)
+            ret[target]["comment"] = "Package {} is already set to be held.".format(
+                target
+            )
 
     if added:
-        __zypper__(root=root).call("al", *added)
+        __zypper__.call("al", *added)
```
meaksh, line 2367: Same as above

salt/modules/zypperpkg.py
vzhestkov: Yup, sorry, just noticed it also. `hold` and `unhold` functions in upstream have no `root`.
Fixed

vzhestkov:
@aplanas fixed, also noticed it, removed by accident. Thanks for noticing

vzhestkov:
@agraul I just tried to align the change with https://github.com/saltstack/salt/pull/56935, but there is no distinct `root` paramiter. If it's better to add it to the function signature I'm ok with it. Are there any recommendations documented about it?

agraul:
> @agraul I just tried to align the change with [saltstack/salt#56935](https://github.com/saltstack/salt/pull/56935), but there is no distinct `root` paramiter. If it's better to add it to the function signature I'm ok with it. Are there any recommendations documented about it?

On the one hand I think "it's a 1:1 backport of an upstream PR" is a valid reason to do things in not-optimal ways, but in this case I think adding `root` as a keyword argument is fine. It's not very likely that there will be further changes upstream that would cause new patches to not apply cleanly. It's also easy to send a small PR upstream that changes all the `kwargs.get("root")` to keyword arguments (not saying you need to do that) and I doubt upstream would decline such a PR.

agraul (approved):

PR #401, opened 2021-07-09
transactional_update: prevent issues with systemd because of chroot errors (bsc#1187787)

meaksh:
### What does this PR do?

This PR prevents the wrong behavior of `systemd_service` module when it's called by "transactional_update" module. The problem is that "systemctl" is raising an error due the chrooted environment, and that confuses Salt completely, leading into inconsistent states reports due the retcode of systemctl:

```
# systemctl is-active osad
Running in chroot, ignoring command 'is-active'
# echo $?
0

# SYSTEMD_IGNORE_CHROOT=1 systemctl is-active osad
inactive
# echo $?
3
```

### DIFF
--- a/salt/modules/transactional_update.py
+++ b/salt/modules/transactional_update.py
@@ -345,7 +345,7 @@ def _pkg_params(pkg, pkgs, args):
 
 def _cmd(cmd, retcode=False):
     """Utility function to run commands."""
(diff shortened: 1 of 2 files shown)


aplanas (approved):

meaksh:
We fixed this problem using a different approach. This PR is not longer needed.

Closing.

PR #405, opened 2021-07-30
Add missing aarch64 to rpm package architectures

vzhestkov:
### What does this PR do?

Fixes false negative results on using `pkg.installed` with architecture specification in package name (ex. `bash.aarch64`)
`zypperpkg` module unable to normalize package name with unknown architecture specified.

### Previous Behavior
False negative result on `pkg.installed` with specifying package name with architecture.

### New Behavior
Normal behavior

### Tests
Yes

### Upstream PR
https://github.com/saltstack/salt/pull/60650

### DIFF
--- a/salt/utils/pkg/rpm.py
+++ b/salt/utils/pkg/rpm.py
@@ -33,7 +33,7 @@ ARCHES_ALPHA = (
     "alphaev68",
     "alphaev7",
 )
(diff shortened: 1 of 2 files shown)


SchoolGuy (approved):
From a code perspective I say this is LGTM. In the Salt logical context I can't tell right now.

meaksh (approved):

PR #408, opened 2021-08-08
fix typo in grains/core.py

Shirocco88:
fix typo

### What does this PR do?
fix typo in grains/core.py

fix typo in grains/core.py

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -87,7 +87,7 @@ if salt.utils.platform.is_windows():
         HAS_WMI = True
     except ImportError:
         log.exception(
-            "Unable to import Python wmi module, some core grains " "will be missing"
+            "Unable to import Python wmi module, some core grains will be missing"
         )
 
 


agraul:
Thanks for catching this @Shirocco88! Upstream recently did a cleanup of these formatting issues (https://github.com/saltstack/salt/pull/60494/commits/0696b6467f47f406b39cd7872a1130bf92abd729) and I wonder if the work required to add a patch for just this one instance is worth it. We could also take upstream's changes with the next version of Salt. This "typo" is purely* code formatting, the logged message is the same.

```
>>> logging.exception(
... "first string " "second string"
... )
ERROR:root:first string second string
NoneType: None
>>> logging.exception(
... "first string second string"
... )
ERROR:root:first string second string
NoneType: None
```

Personally I think it's fine to fix all of these at once with upstream's patch.

*: merging the strings means the interpreter doesn't have to do it at runtime, but the performance impact is probably negligible.

PR #413, opened 2021-08-17
Don't use shell="/sbin/nologin" in requisites

agraul:
Using shell="/sbin/nologin" in an onlyif/unless requisite does not really make sense since the condition can't be run. shell=/sbin/nologin is also a common argument, e.g. for user.present.

Fixes: bsc#1188259
Fixes: https://github.com/SUSE/spacewalk/issues/15382

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -889,9 +889,14 @@ class State:
             cmd_opts[run_cmd_arg] = low_data.get(run_cmd_arg)
 
         if "shell" in low_data:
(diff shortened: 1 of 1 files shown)


vzhestkov (approved):
:+1:

meaksh (approved):

agraul:
We might drop this patch once the root cause is fixed in collaboration with upstream: https://github.com/saltstack/salt/issues/59420

PR #414, opened 2021-08-18
Fix salt ssh issues (bsc#1183446, bsc#1182851)

vzhestkov:
Fixes: https://github.com/SUSE/spacewalk/issues/14307 https://github.com/SUSE/spacewalk/issues/14142

### Previous Behavior
Segfaults in `salt-api` child processes and deadlocks on `logging` and `importlib` modules.

### New Behavior
Normal behavior

### Upstream Salt
https://github.com/saltstack/salt/pull/61009

### DIFF
--- a/salt/_logging/impl.py
+++ b/salt/_logging/impl.py
@@ -288,37 +288,31 @@ class SaltLoggingClass(
         else:
             extra["exc_info_on_loglevel"] = exc_info_on_loglevel
 
-        try:
-            logging._acquireLock()
-            if sys.version_info < (3,):
-                LOGGING_LOGGER_CLASS._log(
-                    self, level, msg, args, exc_info=exc_info, extra=extra
-                )
-            elif sys.version_info < (3, 8):
-                LOGGING_LOGGER_CLASS._log(
-                    self,
-                    level,
-                    msg,
-                    args,
-                    exc_info=exc_info,
-                    extra=extra,
-                    stack_info=stack_info,
-                )
-            else:
-                LOGGING_LOGGER_CLASS._log(
-                    self,
-                    level,
-                    msg,
-                    args,
-                    exc_info=exc_info,
-                    extra=extra,
-                    stack_info=stack_info,
-                    stacklevel=stacklevel,
-                )
-        except:
-            pass
-        finally:
-            logging._releaseLock()
+        if sys.version_info < (3,):
+            LOGGING_LOGGER_CLASS._log(
+                self, level, msg, args, exc_info=exc_info, extra=extra
+            )
+        elif sys.version_info < (3, 8):
+            LOGGING_LOGGER_CLASS._log(
+                self,
+                level,
+                msg,
+                args,
+                exc_info=exc_info,
+                extra=extra,
+                stack_info=stack_info,
+            )
+        else:
+            LOGGING_LOGGER_CLASS._log(
+                self,
+                level,
+                msg,
+                args,
+                exc_info=exc_info,
+                extra=extra,
+                stack_info=stack_info,
+                stacklevel=stacklevel,
+            )
 
     def makeRecord(
         self,
--- a/salt/_logging/impl.py
+++ b/salt/_logging/impl.py
@@ -288,31 +288,37 @@ class SaltLoggingClass(
         else:
             extra["exc_info_on_loglevel"] = exc_info_on_loglevel
 
-        if sys.version_info < (3,):
-            LOGGING_LOGGER_CLASS._log(
-                self, level, msg, args, exc_info=exc_info, extra=extra
-            )
-        elif sys.version_info < (3, 8):
-            LOGGING_LOGGER_CLASS._log(
-                self,
-                level,
-                msg,
-                args,
-                exc_info=exc_info,
-                extra=extra,
-                stack_info=stack_info,
-            )
-        else:
-            LOGGING_LOGGER_CLASS._log(
-                self,
-                level,
-                msg,
-                args,
-                exc_info=exc_info,
-                extra=extra,
-                stack_info=stack_info,
-                stacklevel=stacklevel,
-            )
+        try:
+            logging._acquireLock()
+            if sys.version_info < (3,):
+                LOGGING_LOGGER_CLASS._log(
+                    self, level, msg, args, exc_info=exc_info, extra=extra
+                )
+            elif sys.version_info < (3, 8):
+                LOGGING_LOGGER_CLASS._log(
+                    self,
+                    level,
+                    msg,
+                    args,
+                    exc_info=exc_info,
+                    extra=extra,
+                    stack_info=stack_info,
+                )
+            else:
+                LOGGING_LOGGER_CLASS._log(
+                    self,
+                    level,
+                    msg,
+                    args,
+                    exc_info=exc_info,
+                    extra=extra,
+                    stack_info=stack_info,
+                    stacklevel=stacklevel,
+                )
+        except:
+            pass
+        finally:
+            logging._releaseLock()
 
     def makeRecord(
         self,
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -132,6 +132,10 @@ DEBUG="{{DEBUG}}"
 if [ -n "$DEBUG" ]
     then set -x
 fi
+SALTDIR="{{SALTDIR}}"
+if [ -n "$SALTDIR" ]
+    then find "$SALTDIR" -name 'salt_state*.tgz' -mtime +1 -delete 2>/dev/null || true
+fi
 SET_PATH="{{SET_PATH}}"
 if [ -n "$SET_PATH" ]
     then export PATH={{SET_PATH}}
@@ -903,6 +907,9 @@ class Single:
             self.wipe = False
         else:
             self.wipe = bool(self.opts.get("ssh_wipe"))
+        self.wipe_state = ""
+        if kwargs.get("wipe_state"):
+            self.wipe_state = kwargs["wipe_state"]
         if kwargs.get("thin_dir"):
             self.thin_dir = kwargs["thin_dir"]
         elif self.winrm:
@@ -1315,6 +1322,7 @@ OPTIONS.hashfunc = '{hashfunc}'
 OPTIONS.version = '{version}'
 OPTIONS.ext_mods = '{ext_mods}'
 OPTIONS.wipe = {wipe}
+OPTIONS.wipe_state = '{wipe_state}'
 OPTIONS.tty = {tty}
 OPTIONS.cmd_umask = {cmd_umask}
 OPTIONS.code_checksum = {code_checksum}
@@ -1327,6 +1335,7 @@ ARGS = {arguments}\n'''.format(
             version=salt.version.__version__,
             ext_mods=self.mods.get("version", ""),
             wipe=self.wipe,
+            wipe_state=self.wipe_state,
             tty=self.tty,
             cmd_umask=self.cmd_umask,
             code_checksum=thin_code_digest,
@@ -1342,6 +1351,7 @@ ARGS = {arguments}\n'''.format(
                 SSH_PY_CODE=py_code_enc,
                 HOST_PY_MAJOR=sys.version_info[0],
                 SET_PATH=self.set_path,
+                SALTDIR=self.thin_dir,
             )
         else:
             cmd = saltwinshell.gen_shim(py_code_enc)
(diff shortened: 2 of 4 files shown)


salt/_logging/impl.py
```diff
@@ -288,37 +288,31 @@ def _log(
         else:
             extra["exc_info_on_loglevel"] = exc_info_on_loglevel
 
-        try:
-            logging._acquireLock()
```
agraul, line 292: Do you know why this was here in the first place? Are you certain that we don't add a race condition by removing the lock?

salt/client/ssh/__init__.py
```diff
@@ -537,37 +549,38 @@ def handle_routine(self, que, opts, host, target, mine=False):
...
-        logging._releaseLock()
-        # This job is done, yield
-        try:
-            data = salt.utils.json.find_json(stdout)
-            if len(data) < 2 and "local" in data:
-                ret["ret"] = data["local"]
-            else:
+        ret = {}
+        fake_lock = FakeLock()
+        with patch('importlib._bootstrap._get_module_lock', return_value=fake_lock):
```
agraul, line 554: This seems really hacky, do we really need that? I see you removed the `_acquireLock` & `_releaseLock` calls below, so I assume it's needed for something else. What codepath needs this mocked?

salt/_logging/impl.py
vzhestkov: I tested this change with stress test running for slightly more than 24 hours with no any issues, if I keep this lock acquire and release in place it leads to deadlocks on calling `logging.getLogger(__name__)` in `salt-api` child processes almost randomly in few minutes up to few hours, but finally `salt-api` stops responding on incoming requests and keeps connections open from taskomatic ssh push.
The worst thing that deadlocks on logging and importlib modules with using multiprocessing are known as a bugs for a long time, but still no any straight solution, just the recommendation for logging to process all the log messages with distinct process passing messages for logging through the `multiprocessing.Queue`.

salt/client/ssh/__init__.py
vzhestkov: yes, it's hacky, but it's the only way I found to solve importlib deadlocks (no any side effects found in 24 hours stress test), it's not about `_acquireLock` and `_releaseLock` for `logging`. `importlib._bootstrap._get_module_lock` is returning individual lock objects for each module and in multiprocessing case it also leads to deadlocks, there are some changes for Python 3.8, but I tested salt in Tumbleweed with Python 3.8 and it seems that deadlocks with `importlib` are still there in Python 3.8

salt/_logging/impl.py
agraul: I guess without the locks the logs will be out of order or what's the original purpose behind them?
I just want to make sure we understand what we're changing. Could you share how you ran the stress test and its scope?

salt/client/ssh/__init__.py
agraul: >  importlib._bootstrap._get_module_lock is returning individual lock objects for each module and in multiprocessing case it also leads to deadlocks

Can you elaborate on this part? I don't understand the cause of the problem yet. I hope we find a cleaner solution than monkey patching part of the standard library in the code we ship to our customers.

salt/_logging/impl.py
vzhestkov: It runs a lot of simulataneous events with state.apply on salt-ssh systems. The state is the following:
```
dummy-test:
  cmd.run:
  - name: for (( i=1; i<=5; i++ )); do  printf "DUMMYOUTPUT"; sleep 1; done
```
The state apply should take some time and multiple simultaneous ssh sessions causing the issues for the customer. I tried to reproduce the issues in amplified syntetic form.

salt/client/ssh/__init__.py
vzhestkov: Here is the example of traceback for salt-api child process get stuck on `acquire`:
```
Traceback (most recent call first):
  File "<frozen importlib._bootstrap>", line 98, in acquire
  File "<frozen importlib._bootstrap>", line 149, in __enter__
  File "<frozen importlib._bootstrap>", line 586, in _exec
  File "<frozen importlib._bootstrap>", line 262, in _load_module_shim
  File "<frozen importlib._bootstrap_external>", line 852, in load_module
  File "<frozen importlib._bootstrap_external>", line 1027, in load_module
  File "<frozen importlib._bootstrap_external>", line 1034, in _check_name_wrapper
  File "/usr/lib/python3.8/site-packages/salt/loader.py", line 4779, in _load_module
  File "/usr/lib/python3.8/site-packages/salt/loader.py", line 1926, in _inner_load
    if self._load_module(name) and key in self._dict:
  File "/usr/lib/python3.8/site-packages/salt/loader.py", line 2193, in _load
  File "/usr/lib/python3.8/site-packages/salt/utils/lazy.py", line 99, in __getitem__
...
```
It could fail on loading different modules. I have not found any common pattern for it. You can take a look on the sources for lock object used in importlib and it seems it's not intended to be used with multiprocessing. And the bad thing that it's happening in frozen module.

vzhestkov: The example above was captured from Tumbleweed with Python 3.8 which still has the same issues with logging and importlib. There was a quite strange fix for importlib which most probably did not help at all: https://github.com/python/cpython/pull/17518

vzhestkov:
@meaksh @agraul I've reverted one of the commits and left here only the proper fixes which can be promoted to upstream also.
Please review it and I'll port it to `3003.3` also

salt/client/ssh/__init__.py
```diff
@@ -903,6 +905,9 @@ def __init__(
             self.wipe = False
         else:
             self.wipe = bool(self.opts.get("ssh_wipe"))
+        self.wipe_state = ""
```
agraul, line 908: `wipe_state` is not a name that makes it obvious how this is used. I would expect something states such as `started`, `running`, 'finished' or sth like that, but that does not seem to be what this variable is actually for. Can we come up wtih a better name?

salt/client/ssh/ssh_py_shim.py
```diff
@@ -273,6 +274,11 @@ def main(argv):  # pylint: disable=W0613
     """
     Main program body
     """
+    try:
+        lock_fh = open("{0}.lock".format(OPTIONS.saltdir), "w")
+        fcntl.flock(lock_fh.fileno(), fcntl.LOCK_EX)
```
agraul, line 279: Can you please leave comments in the code that answer the two questions below?

1) what is the problem that is fixed by adding this lock
2) how is adding the lock fixing the problem


By the way, you don't need to call `.fileno()` if you don't want to :)

salt/client/ssh/ssh_py_shim.py
```diff
@@ -273,6 +274,11 @@ def main(argv):  # pylint: disable=W0613
...
+    except (IOError, OSError):
```
agraul, line 280: This is Python3.6+ code only, we only need to care about `OSError`. Since we're just `pass`ing, we could also use the following. Just learned about that recently, we don't have to start using that **now**.

```python
with contextlib.suppress(OSError):
    code_that_might_raise_os_error()
```

agraul (changes requested):
I can't speak to how well this works, but I know you've tested it.

 I would like to clarify the code a bit so that it is easier for me to understand what problem you have fixed and in what way. Can we use better self-explanatory names and leave some comments about what you're doing? Having them will make it easier for all of us to understand what's going on in the future when we have to re-visit this code.

From the PR description it is also not possible for me to understand the problem. It would help me with my review if there were more information in the PR. For example, where do the segfaults and deadlocks come from? How is your change avoiding those circumstances?

salt/client/ssh/__init__.py
vzhestkov: `wipe_state` stands for wipe the state file after processing it. As the file name is not unique we need to clean up the files after processing them. Before the file name was static and overwritten each time it processes the state.
Is `wipe_state_file` more clear for such a case?

salt/client/ssh/ssh_py_shim.py
vzhestkov: ssh_py_shim is working on salt ssh system side

agraul: Right, but Salt 3002.2 is only used by SSH clients that run Python3.6+. Python2 SSH clients don't use the code from this branch.

salt/client/ssh/__init__.py
agraul: I see. Yes, I think `wipe_state_file` is clearer for the name of the variable that holds the name of the state file that's getting wiped.

salt/client/ssh/ssh_py_shim.py
vzhestkov: I think we are mixing diffetent things in this discussion.
This exact `ssh_py_shim.py` file is used by salt-api to prepare the script passed in base64 to the shell script by ssh to the managed system, so this code is used for SLE11 also and I tested it on it.

vzhestkov: `fileno()` call removed and comments appended with https://github.com/openSUSE/salt/pull/414/commits/89bab932fb7307a501b5ba0bee6f959a8c1973ae
Are you ok with the rest part? The code in this part must be Python 2.6 compatible

salt/client/ssh/ssh_py_shim.py
vzhestkov: `fileno()` call removed and comments appended with https://github.com/openSUSE/salt/pull/414/commits/4e54a909c856e0852cb8dbc08f6fcc8e9a3c66ed

salt/client/ssh/__init__.py
vzhestkov: `wipe_state` was renamed to `wipe_state_file`  in https://github.com/openSUSE/salt/pull/414/commits/89bab932fb7307a501b5ba0bee6f959a8c1973ae

salt/client/ssh/ssh_py_shim.py
agraul: Thanks for the explanation, you're right this file is special in that it is used on all ssh clients, even those that get a different salt codebase as their execution environment.

salt/client/ssh/ssh_py_shim.py
```diff
@@ -395,6 +404,11 @@ def main(argv):  # pylint: disable=W0613
         shutil.rmtree(OPTIONS.saltdir)
     else:
         retcode = subprocess.call(salt_argv)
+    if OPTIONS.wipe_state_file:
+        try:
+            os.remove(os.path.join(OPTIONS.saltdir, OPTIONS.wipe_state_file))
+        except OSError:
```
agraul, line 410: Given our discussion above, I think this needs to be 
```suggestion
        except (IOError, OSError):
```
as well.

salt/client/ssh/ssh_py_shim.py
vzhestkov: `IOError` and `OSError` should be used both for handling the lock as Python2 producing IOErrror while Python3 - OSError, but in case of calling `os.remove` Python2 and Python3 producing OSError

salt/client/ssh/ssh_py_shim.py
agraul: Thanks for the comments. It's definitely better now

salt/client/ssh/ssh_py_shim.py
agraul: Interesting, my lack of Python2 knowledge shows again. Thanks for the clarification!

agraul (commented):
Again, haven't tested it and I don't fully understand the bug from the pull request, but the questions from earlier are resolved. The changes themselves are "ok", but I can't speak to what problem they are actually fixing.

 It would really help pull request reviews to have the context needed to understand bug fixes (i.e. what are the symptoms of a bug, what is the cause of the bug, how is this change fixing the cause). I don't expect a copy of the bug report, but a summary would be nice.

vzhestkov:
There are two distinct issues fixed, but both related to salt-ssh.
1. In case of opening simultaneous session when both of them require updating salt-thin, one of the sessions rewrites some of the files, causing tracebacks. Flock in the shim helped in this situation and anyway we need to serialize state applies as it could cause some issues also
2. Also in case of simultaneous sessions salt-api could rewrite salt_state.tgz and cause blank result for one of the sessions (no tracebacks, total silence, but just empty result).

Both look similar and you might think that only solution with the lock could help, but these issues could happen independently as processed with different ssh/scp commands and locking is not preventing rewriting of the state tarball.

vzhestkov:
PR to the upstream: https://github.com/saltstack/salt/pull/61009

vzhestkov:
Closing as better solution was implemented with salt-ssh with salt bundle change.

PR #415, opened 2021-08-19
Prevent pkg plugins errors on missing cookie path (bsc#1186738) - 3002.2

vzhestkov:
Fixes: https://github.com/SUSE/spacewalk/issues/15079

### Previous Behavior
Exceptions in case if salt-minion cache path does not exist yet, mosty relevant for Debian/Ubuntu as cache path is created only on salt-minion start.

### New Behavior
No exceptions.

### DIFF
--- a/scripts/suse/dpkg/dpkgnotify
+++ b/scripts/suse/dpkg/dpkgnotify
@@ -35,9 +35,19 @@ def dpkg_post_invoke():
     """
     Hook after the package installation transaction.
     """
-    if 'SALT_RUNNING' not in os.environ:
-        with open(CK_PATH, 'w') as ck_fh:
-            ck_fh.write('{chksum} {mtime}\n'.format(chksum=_get_checksum(), mtime=_get_mtime()))
+    if "SALT_RUNNING" not in os.environ:
+        try:
+            ck_dir = os.path.dirname(CK_PATH)
+            if not os.path.exists(ck_dir):
+                os.makedirs(ck_dir)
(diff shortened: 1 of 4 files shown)


scripts/suse/dpkg/dpkgnotify
```diff
@@ -35,9 +35,19 @@ def dpkg_post_invoke():
...
+            if not os.path.exists(ck_dir):
+                os.makedirs(ck_dir)
+            with open(CK_PATH, "w") as ck_fh:
+                ck_fh.write(
+                    "{chksum} {mtime}\n".format(
+                        chksum=_get_checksum(), mtime=_get_mtime()
+                    )
+                )
+        except Exception:  # pylint: disable=broad-except:
+            pass
```
agraul, line 50: what exception is caught here? Please be explicit.

vzhestkov: Depending on python version it could be different `FileNotFoundError`or `IOError`, I think too much effort to catch the exception which actually shouldn't be passed to the user

vzhestkov: Exception handling narrowed down with https://github.com/openSUSE/salt/pull/415/commits/ef523ebaf21087314e31ecd904bdacde0077f2bb
For `Zypper` the message goes to `/var/log/zypper.log`, for `APT` to `stderr`, but `Yum` seems not calling plugins at all, at least on CentOS 8

agraul (approved):

PR #418, opened 2021-08-30
templates: move the globals up to the Environment (Jinja2 3.0.0)

aplanas:
### What does this PR do?
When creating a Jinja2 environment, we populate the globals in the
Template object that we generate from the environment.  This cause a
problem when there is a {% include "./file.sls" %} in the template, as
cannot find in the environment globals information like the "tpldir",
for example, making the relative path to be unresolved.

Seems that in Jinja2 2.X this behaviour is not present, so attaching the
globals to the Template will make the include to work, but since Jinja2
3.0.0 this is not the case.  Maybe related with the re-architecture from
pallets/jinja#295

This patch populate the globals in the Environment level, making this
and other variables reachable by the Jinja templates.

Also fix a wrong call to TemplateNotFound

Fixes: #55159

- [ ] Docs
- [X] Tests written/updated

### Commits signed with GPG?
Yes

Upstream: https://github.com/saltstack/salt/pull/60811

### DIFF
--- a/tests/unit/utils/test_jinja.py
+++ b/tests/unit/utils/test_jinja.py
@@ -612,6 +612,22 @@ class TestGetTemplate(TestCase):
             dict(opts=self.local_opts, saltenv="test", salt=self.local_salt),
         )
 
+    def test_relative_include(self):
+        template = "{% include './hello_import' %}"
+        expected = "Hey world !a b !"
+        filename = os.path.join(self.template_dir, "hello_import")
+        with salt.utils.files.fopen(filename) as fp_:
+            out = render_jinja_tmpl(
+                template,
+                dict(
+                    opts=self.local_opts,
+                    saltenv="test",
+                    salt=self.local_salt,
+                    tpldir=self.template_dir,
+                ),
+            )
+        self.assertEqual(out, expected)
+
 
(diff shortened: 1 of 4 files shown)


tests/unit/utils/test_jinja.py
```diff
@@ -612,6 +612,22 @@ def test_render_with_undefined_variable_unicode(self):
             dict(opts=self.local_opts, saltenv="test", salt=self.local_salt),
         )
 
+    def test_relative_include(self):
```
vzhestkov, line 615: Is this test really make sense? I see it's not failing even without thix fix applied.

vzhestkov (commented):
I'm ok to accept it and merge, but probably the test could be more relevant for it?

tests/unit/utils/test_jinja.py
aplanas: IMHO makes sense. You can reproduce the fail when installing jinja 3.0.1 that is in TW. This test fails with this Jinja, and succeed with 2.11.

vzhestkov: Oh, sorry. You are right. Missed that it's dependant on Jinja2 vertion used.

vzhestkov (approved):
:+1:

PR #421, opened 2021-09-03
Fix the regression brought in with PR#402

vzhestkov:
### What does this PR do?

Fixes regression brought in with https://github.com/openSUSE/salt/pull/402


agraul:
There should be two more regressions: https://github.com/saltstack/salt/pull/57825 and subsequently https://github.com/openSUSE/salt/pull/413/ introduced by the consolidation.

PR #427, opened 2021-09-30
Fix file.find tracebacks with non utf8 file names (bsc#1190114)

vzhestkov:
Fixes: https://github.com/SUSE/spacewalk/issues/15792

### Previous Behavior
Tracebacks if `file.find` called for firectory containing non utf8 characters in the file names inside directory.
`3002.2` has no such issue

### New Behavior
Normal output on calling `file.find`.

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -1078,6 +1078,10 @@ def find(path, *args, **kwargs):
         return 'error: {0}'.format(ex)
 
     ret = [item for i in [finder.find(p) for p in glob.glob(os.path.expanduser(path))] for item in i]
+    if six.PY2:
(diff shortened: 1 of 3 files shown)


vzhestkov:
No, `3002.2` and higher have no such issue. I didn't find any upstream PR for it.

agraul:
> No, 3002.2 and higher have no such issue.

That's surprising, the you've touched code is the same between 3002.2 and 3000. Could it be that this problem only exists in Python2?

vzhestkov:
I also supposed so, but I've tested both python2-salt and python3-salt on sle12 with salt 3000 and both producing the tracebacks, but different ones, so the fix contains conditional parts, but there are no tracebacks on sle15 with salt 3002.2

PR #438, opened 2021-10-06
[3000] sql json - realign with upstream

cbosdo:
Add the missing `versionadded` added upstream after the previous PR

### DIFF
--- a/salt/pillar/sql_base.py
+++ b/salt/pillar/sql_base.py
@@ -148,6 +148,9 @@ It is capable of handling single rows or multiple rows per minion.
 
 Configuration of the connection depends on the adapter in use.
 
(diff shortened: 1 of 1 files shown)


cbosdo:
@meaksh you probably want to look at this PR and its ports

meaksh:
@cbosdo I am not really sure that we want to push this patch to our package, as it will generate a new patch to carry and maintain in our package simply because of this `versionadded` missing change.

Is it a problem to live without this one? Or at least until we have something else to push to this same file?

What do you think?

cbosdo:
It's up to you: those PRs were only to sync the changes of upstream PR with the ones that was merged in our package. If you think it's fine without it let's close the PRs without merging

PR #448, opened 2021-10-25
Add rpm_vercmp python library for version comparison - 3003.3

vzhestkov:
### What does this PR do?

Backport of https://github.com/saltstack/salt/pull/60815

Helps to avoid `python-rpm` binary dependency for Salt Bundle

Requires this PR to handle `^` the proper way: https://github.com/sassoftware/python-rpm-vercmp/pull/3


tests/unit/modules/test_rpm_lowpkg.py
```diff
@@ -1,478 +0,0 @@
...
-            self.assertFalse(_called_with_root(mock))
-
-    def test_remove_gpg_key(self):
-        """
-        Test remove_gpg_key
-        """
-        mock = MagicMock(return_value=0)
-        with patch.dict(rpm.__salt__, {"cmd.retcode": mock}):
-            self.assertTrue(rpm.remove_gpg_key("gpg-pubkey-1"))
-            self.assertFalse(_called_with_root(mock))
```
meaksh, line 478: I think we should move these tests to the new pytests file. These are not included in the pytest file from the upstream PR, as these tests are probably coming from one of our downstream patches for 3002.2.

vzhestkov: You mean the tests related to gpg, right? I'll move them.

meaksh: Yeah, exactly. Thanks! :+1:

vzhestkov: I just pushed https://github.com/openSUSE/salt/pull/448/commits/d9a0120c61f49b4f8f09c21facb7859e8f237dc5 moving the GPG tests back and realigns them to pytest

PR #450, opened 2021-11-03
dnfnotify pkgset plugin implementation - 3002.2

vzhestkov:
### What does this PR do?

Implements `dnfnotify` plugin for `pkgset` beacon.
`yumnotify` is not working with DNF as dnf itself is not invoking Yum plugins.

### Previous Behavior
Cookie file is not updated on installed packages list change

### New Behavior
`pkgset` works the same way as Zypper and Yum notify plugins

### Upstream PR
https://github.com/saltstack/salt/pull/61179

### DIFF
--- a/scripts/suse/dnf/plugins/README.md
+++ b/scripts/suse/dnf/plugins/README.md
@@ -1,0 +1,21 @@
+## What it is
+
+Plugin which provides a notification mechanism to Salt, if DNF is
+used outside of it.
+
+## Installation
+
+Configuration files are going to:
+
+	`/etc/dnf/plugins/[name].conf`
+
+Plugin itself goes to:
+
+	`%{python_sitelib}/dnf-plugins/[name].py`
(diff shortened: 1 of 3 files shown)


meaksh (approved):

SchoolGuy:
@vzhestkov Since we add new functionality, could we maybe add (a) test(s) for it?

vzhestkov:
@SchoolGuy the proper test is not possible here, but with mocking almost everything the test becomes useless. I think this is a reason why we have no tests for any of these plugins.

SchoolGuy:
@vzhestkov So we can't even setup an integration test where we install vim and see if we get a notification?

vzhestkov:
Plugin is acually is not a part of salt, but should be run by dnf in this case, so building a proper test will take much more effort than plugin itself, it's better to check it with Cucumber in Uyuni.

SchoolGuy (approved):
LGTM

agraul (approved):

PR #463, opened 2021-11-26
3003.3 event send fix 

cbosdo:
Port of https://github.com/saltstack/salt/pull/61256


cbosdo:
closing since 3004 is the new thing

PR #464, opened 2021-12-13
Fix possible traceback on ip6_interface grain (bsc#1193565) - 3000

vzhestkov:
### What does this PR do?
Fixes possible traceback on `ip6_interface` at `grains.core` in case of having multiple IP addresses assigned.
Python 2 has `socket.error` which is not inherited from `OSError` as in Python 3

Fixes: https://github.com/SUSE/spacewalk/issues/16516
Appears in: https://github.com/openSUSE/salt/commit/a5fb276e3a925f6dab4f97e3df9163c72c9818b2

### Previous Behavior
Possible traceback on `ip6_interface` at `grain.core` in case of having multiple IP addresses assigned.

### New Behavior
Handling exception with no traceback.

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -2377,7 +2377,7 @@ def ip6_interfaces():
                 try:
                     socket.inet_pton(socket.AF_INET6, secondary["address"])
                     iface_ips.append(secondary["address"])
-                except OSError:
+                except (OSError, socket.error):
                     pass
         ret[face] = iface_ips
     return {'ip6_interfaces': ret}


meaksh (approved):

meaksh:
Thanks @vzhestkov :+1: 

I'm pushing this to the upstream PR, which is still open: https://github.com/saltstack/salt/pull/61061

meaksh:
Well, in fact this change is not needed in upstream "master" branch as it is only affecting Python 2, so I'm not including it on the PR.

If you think I should them, just tell me and I will add the commit :)

vzhestkov:
@meaksh yes, no need to upstream it as it's relevant for Python 2 only

brett060102 (approved):
I am OK with the change, but is the failing CI test normal? I does not look lie it would be realted to this change, but I can't be sure.

vzhestkov:
@brett060102 yes, this failing test is totally unrelated to the change.

PR #467, opened 2021-12-15
Fix salt-minion.service shown as reloading after calling apache2ctl (bsc#1193357) - 3002.2

vzhestkov:
### What does this PR do?

For some reason on calling `apache2ctl` from `salt-minion` with some `apache` module functions under systemd service, the service shown as reloading, it doesn't affect the `salt-minion` process, but only the service status shown.

The root cause is the side effect on changing some environment variables with `apache2ctl`.

Upstream PR: https://github.com/saltstack/salt/pull/61375

Fixes: https://github.com/SUSE/spacewalk/issues/16494

### Previous Behavior
`salt-minion` shown as `reloading` forewer after calling `apache2ctl` with `apache.modules` and some other `apache` module functions

### New Behavior
Normal systemd service behavior

### DIFF
--- a/salt/modules/apache.py
+++ b/salt/modules/apache.py
@@ -19,6 +19,7 @@ from __future__ import (
 )
 
 import logging
+import os
 import re
 
 # Import salt libs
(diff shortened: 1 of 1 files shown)


SchoolGuy (approved):
This is tested and thus I believe it is the most sane way to approach this. Did we upstream this or is it just a backport from upstrem?

vzhestkov:
I didn't find upstream issue for it. I think it's better to promote to upstream. Will do it.

vzhestkov:
Closing in favor of https://github.com/openSUSE/salt/pull/472

PR #468, opened 2021-12-15
Fix salt-minion.service shown as reloading after calling apache2ctl (bsc#1193357) - 3000

vzhestkov:
### What does this PR do?

Backport of https://github.com/openSUSE/salt/pull/467 to `3000`.
No need to port it to `3000.3` as it's not used under systemd unit, but with salt-ssh only.

For some reason on calling `apache2ctl` from `salt-minion` with some `apache` module functions under systemd service, the service shown as reloading, it doesn't affect the `salt-minion` process, but only the service status shown.

The root cause is the side effect on changing some environment variables with `apache2ctl`.

Fixes: https://github.com/SUSE/spacewalk/issues/16494

### Previous Behavior
`salt-minion` shown as `reloading` forewer after calling `apache2ctl` with `apache.modules` and some other `apache` module functions

### New Behavior
Normal systemd service behavior

### DIFF
--- a/salt/modules/apache.py
+++ b/salt/modules/apache.py
@@ -11,6 +11,7 @@ Support for Apache
 
 # Import python libs
 from __future__ import absolute_import, generators, print_function, with_statement, unicode_literals
+import os
 import re
 import logging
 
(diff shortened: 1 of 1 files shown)


vzhestkov:
Closing in favor of https://github.com/openSUSE/salt/pull/474

PR #475, opened 2022-01-19
Add all ssh kwargs to sanitize_kwargs method - 3002.2

vzhestkov:
### What does this PR do?

Backport of https://github.com/saltstack/salt/pull/59748 from upstream

### Previous Behavior
Some significant ssh options are wiping from the request for salt ssh session.
In our particular case we need `roster` parameter to be passed to `salt.client.ssh.SSH` with `salt-api`.

### New Behavior
Passing ssh options to `salt.client.ssh.SSH`

### Porting
`3003.3` and higher versions already contain this fix, so no need to port it.


vzhestkov:
@meaksh not really sure about removing the lock, but it causing deadlock on testing and I also see the deadlocks related to this particular lock on the tricky L3 about salt-ssh issues.

meaksh (approved):
Ok, let's give it a try

PR #476, opened 2022-01-21
Add salt-ssh support with venv-salt-minion - 3002.2

vzhestkov:
### What does this PR do?

Adds support of `venv-salt-minion` to handle salt-ssh session on the minion.
Extends ssh pre flight script feature with passing additional parameter `ssh_pre_flight_args` from the roster.

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -245,7 +245,9 @@ class SSH:
             else "glob"
         )
         self._expand_target()
-        self.roster = salt.roster.Roster(self.opts, self.opts.get("roster", "flat"), context=context)
+        self.roster = salt.roster.Roster(
+            self.opts, self.opts.get("roster", "flat"), context=context
+        )
         self.targets = self.roster.targets(self.opts["tgt"], self.tgt_type)
         if not self.targets:
             self._update_targets()
@@ -622,7 +624,10 @@ class SSH:
                             )
                             if (
                                 pid_running and prev_session_running < self.max_pid_wait
-                            ) or (not pid_running and prev_session_running < self.ssh_session_grace_time):
+                            ) or (
+                                not pid_running
+                                and prev_session_running < self.ssh_session_grace_time
+                            ):
                                 targets_queue.append(host)
                                 time.sleep(0.3)
                                 continue
@@ -669,11 +674,15 @@ class SSH:
                 routine.start()
                 running[host] = {"thread": routine}
                 with salt.utils.files.flopen(self.session_flock_file, "w"):
-                    self.cache.store("salt-ssh/session", host, {
-                        "pid": routine.pid,
-                        "master_id": self.master_id,
-                        "ts": time.time(),
-                    })
+                    self.cache.store(
+                        "salt-ssh/session",
+                        host,
+                        {
+                            "pid": routine.pid,
+                            "master_id": self.master_id,
+                            "ts": time.time(),
+                        },
+                    )
                 continue
             ret = {}
             try:
@@ -715,11 +724,15 @@ class SSH:
                 if host in running:
                     running.pop(host)
                     with salt.utils.files.flopen(self.session_flock_file, "w"):
-                        self.cache.store("salt-ssh/session", host, {
-                            "pid": 0,
-                            "master_id": self.master_id,
-                            "ts": time.time(),
-                        })
+                        self.cache.store(
+                            "salt-ssh/session",
+                            host,
+                            {
+                                "pid": 0,
+                                "master_id": self.master_id,
+                                "ts": time.time(),
+                            },
+                        )
             if len(rets) >= len(self.targets):
                 break
             # Sleep when limit or all threads started
@@ -1114,7 +1127,8 @@ class Single:
         """
         if self.mods.get("file"):
             self.shell.send(
-                self.mods["file"], os.path.join(self.thin_dir, "salt-ext_mods.tgz"),
+                self.mods["file"],
+                os.path.join(self.thin_dir, "salt-ext_mods.tgz"),
             )
         return True
 
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -555,6 +555,7 @@ class SSH:
         """
         Run the routine in a "Thread", put a dict on the queue
         """
+        salt.loader.LOAD_LOCK.release()
         opts = copy.deepcopy(opts)
         single = Single(
             opts,
@@ -671,7 +672,11 @@ class SSH:
                 # Explicitly call garbage collector to prevent possible segfault
                 # in salt-api child process. (bsc#1188607)
                 gc.collect()
-                routine.start()
+                try:
+                    salt.loader.LOAD_LOCK.acquire()
+                    routine.start()
+                finally:
+                    salt.loader.LOAD_LOCK.release()
                 running[host] = {"thread": routine}
                 with salt.utils.files.flopen(self.session_flock_file, "w"):
                     self.cache.store(
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -45,6 +45,7 @@ import salt.utils.stringutils
 import salt.utils.thin
 import salt.utils.url
 import salt.utils.verify
+from salt._logging.impl import LOG_LOCK
 from salt.ext import six
 from salt.ext.six.moves import input  # pylint: disable=import-error,redefined-builtin
 from salt.template import compile_template
@@ -555,6 +556,7 @@ class SSH:
         """
         Run the routine in a "Thread", put a dict on the queue
         """
+        LOG_LOCK.release()
         salt.loader.LOAD_LOCK.release()
         opts = copy.deepcopy(opts)
         single = Single(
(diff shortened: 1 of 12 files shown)


salt/client/ssh/ssh_py_shim.py
```diff
@@ -273,55 +273,70 @@ def main(argv):  # pylint: disable=W0613
...
+    if virt_env and "venv-salt-minion" in virt_env:
+        venv_salt_call = os.path.join(virt_env, "bin", "salt-call")
+        if not os.path.exists(venv_salt_call):
+            venv_salt_call = None
+        elif not os.path.exists(OPTIONS.saltdir):
+            os.makedirs(OPTIONS.saltdir)
+            cache_dir = os.path.join(OPTIONS.saltdir, "running_data", "var", "cache")
+            os.makedirs(os.path.join(cache_dir, "salt"))
+            os.symlink("salt", os.path.relpath(os.path.join(cache_dir, "venv-salt-minion")))
+            #os.symlink(os.path.join(cache_dir, "salt"), os.path.relpath(os.path.join(cache_dir, "venv-salt-minion")))
```
meaksh, line 288: Should we remove this line?

```suggestion
```

salt/client/ssh/ssh_py_shim.py
```diff
@@ -273,55 +273,70 @@ def main(argv):  # pylint: disable=W0613
...
-        code_checksum_path = os.path.normpath(
-            os.path.join(OPTIONS.saltdir, "code-checksum")
-        )
-        if not os.path.exists(code_checksum_path) or not os.path.isfile(
-            code_checksum_path
-        ):
-            sys.stderr.write(
-                "WARNING: Unable to locate current code checksum: {0}.\n".format(
-                    code_checksum_path
+    virt_env = os.getenv("VIRTUAL_ENV", None)
```
meaksh, line 277: As mentioned, maybe a small comment to indicate where this `VIRTUAL_ENV` env variable comes from.

salt/client/ssh/ssh_py_shim.py
vzhestkov: sure  :+1: , forgot to remove it

salt/client/ssh/__init__.py
```diff
@@ -1029,13 +1034,26 @@ def check_thin_dir(self):
...
+    def check_venv_hash_file(self):
+        """
+        check if the venv exists on the remote machine
+        """
+        stdout, stderr, retcode = self.shell.exec_cmd(
+            "test -f {}".format(self.venv_hash_file)
+        )
+        if retcode != 0:
+            return False
+        return True
```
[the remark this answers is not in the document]
vzhestkov: Where is my facepalm smiley? I didn't noticed it as it was a copy of `check_thin_dir` and I've just modified the command but missed the rest part. Nice catch :+1: But what to do with `check_thin_dir` in such case? :-D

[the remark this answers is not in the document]
vzhestkov: Fixed in https://github.com/openSUSE/salt/pull/476/commits/486cb813501c1d55c6890ae5a856abc4348ee1ba
Thanks for pointing :+1:

salt/client/ssh/__init__.py
```diff
@@ -1029,13 +1034,24 @@ def check_thin_dir(self):
...
         Deploy salt-thin
         """
-        self.shell.send(
-            self.thin, os.path.join(self.thin_dir, "salt-thin.tgz"),
-        )
+        if not self.check_venv_hash_file():
+            self.shell.send(
+                self.thin,
+                os.path.join(self.thin_dir, "salt-thin.tgz"),
+            )
```
agraul, line 1054: Does the check have to happen inside `deploy()` or can we just not call `deploy()` for systems that get `venv-salt-minion`? I'm not too deep into the salt-thin deployment code right now.

salt/client/ssh/ssh_py_shim.py
```diff
@@ -273,55 +273,72 @@ def main(argv):  # pylint: disable=W0613
...
+        if not os.path.exists(venv_salt_call):
+            venv_salt_call = None
+        elif not os.path.exists(OPTIONS.saltdir):
+            os.makedirs(OPTIONS.saltdir)
+            cache_dir = os.path.join(OPTIONS.saltdir, "running_data", "var", "cache")
+            os.makedirs(os.path.join(cache_dir, "salt"))
+            os.symlink("salt", os.path.relpath(os.path.join(cache_dir, "venv-salt-minion")))
+
+    if venv_salt_call is None:
+        # Use Salt thin only if Salt Bundle (venv-salt-minion) is not available
```
agraul, line 292: I think this code is not up to date with our most recent discussion (https://github.com/uyuni-project/uyuni-rfc/pull/65#discussion_r791614996 and following), am I mistaken or do we need to change it?

agraul (changes requested):
Please correct me if I'm wrong, but I think the code needs a small update after our latest discussion in the Uyuni RFC (https://github.com/uyuni-project/uyuni-rfc/pull/65#discussion_r791614996)

salt/client/ssh/__init__.py
vzhestkov: Actually both questions are interconnected. I mean this one and the following. It depends on the result of running the pre flight script. Initially I tested it with returning success in all of the cases, but as we decided to support only Salt Bundle method of handling the events on the minion, the pre flight script will fail if it's not possible to deploy the Bundle. So it's safe to use this logic and just adjust the pre flight script to our needs. The pre flight script is not a part of any PR yet. Will be included in the separate PR after merging https://github.com/uyuni-project/uyuni/pull/4457.
But this PR is required to be merged first as https://github.com/uyuni-project/uyuni/pull/4457 can't work without this one.

salt/client/ssh/ssh_py_shim.py
vzhestkov: The answer in the previous comment as this points are related.

salt/client/ssh/__init__.py
vzhestkov: Ouch, sorry for the confusion https://github.com/uyuni-project/uyuni/pull/4457 requires not this one but https://github.com/openSUSE/salt/pull/475

meaksh:
Small note here besides of the review.

At some point, we should think into pushing this PR also upstream so we don't need to keep carrying this patch and backporting it across Salt versions.

We should take care of not breaking the default behavior of upstream Salt and making the code to work in our bundle scenario.

vzhestkov:
In current state of the change it only adds extra check if the hash for venv-salt-minion is present on the system, so the logic from upstream version changed in this part only. The overal functionality is changing once the minion really gets the bundle.

salt/client/ssh/__init__.py
```diff
@@ -582,11 +596,40 @@ def handle_ssh(self, mine=False):
...
+                            == self.master_id
+                        ):
+                            pid_running = (
+                                False
+                                if cached_session["pid"] == 0
+                                else psutil.pid_exists(cached_session["pid"])
+                            )
+                            if (
+                                pid_running and prev_session_running < self.max_pid_wait
+                            ) or (not pid_running and prev_session_running < 3):
```
meaksh, line 620: Should we make this grace time of `3` configurable somehow?

salt/client/ssh/__init__.py
```diff
@@ -582,11 +596,40 @@ def handle_ssh(self, mine=False):
...
             if len(running) < self.opts.get("ssh_max_procs", 25) and not init:
-                try:
-                    host = next(target_iter)
-                except StopIteration:
+                if targets_queue:
+                    host = targets_queue.pop(0)
+                else:
                     init = True
                     continue
+                with salt.utils.files.flopen(self.session_flock_file, "w"):
```
meaksh, line 604: What would happen here is the lock file exists already? Isn't this raising an exception?

salt/client/ssh/__init__.py
```diff
@@ -614,15 +657,18 @@ def handle_ssh(self, mine=False):
...
-                    pass
-                finally:
-                    logging._releaseLock()
+                routine = Process(target=self.handle_routine, args=args)
+                # Explicitly call garbage collector to prevent possible segfault
+                # in salt-api child process. (bsc#1188607)
+                gc.collect()
+                routine.start()
                 running[host] = {"thread": routine}
+                with salt.utils.files.flopen(self.session_flock_file, "w"):
```
meaksh, line 666: Same question here, isn't this breaking in case lock file is there already?

salt/roster/__init__.py
```diff
@@ -71,9 +71,14 @@ def __init__(self, opts, backends="flat"):
             self.backends = backends
         if not backends:
             self.backends = ["flat"]
-        utils = salt.loader.utils(self.opts)
-        runner = salt.loader.runner(self.opts, utils=utils)
-        self.rosters = salt.loader.roster(self.opts, runner=runner, utils=utils)
+        with salt.utils.files.flopen(
```
meaksh, line 74: Same here

salt/client/ssh/__init__.py
vzhestkov: Sure, we can. Maybe not very useful, but it could help if the environment has some network issues. Without this waiting the chances to get error with unavailability to open port forwarding are much higher. It seems sshd on the minion side need some time to bring the port down

salt/client/ssh/__init__.py
vzhestkov: It doesn't matter if the file exists already, after implementing the change file exists almost olways. Only posibility to call `flock` inside `salt.utils.files.flopen` makes sense. `salt.utils.files.flopen` is used as it's context aware.

salt/client/ssh/__init__.py
vzhestkov: `salt.utils.files.flopen` used as an atomic operation to properly put the correct data to the cache file with preventing race conditions

salt/client/ssh/__init__.py
vzhestkov: Made configurable with https://github.com/openSUSE/salt/pull/476/commits/5ef5d28051030f2607f78d8f17d359bf66bd6dc0

meaksh (approved):

PR #480, opened 2022-01-27
Fix inspector module export function (bsc#1097531) - 3002.2

vzhestkov:
### What does this PR do?

Port of https://github.com/openSUSE/salt/pull/478 to `3002.2` with adjusting `gzip.open` function calls to Python 3

Upstream PR: https://github.com/saltstack/salt/pull/61530

Fixes the tracebacks on calling `inspector.export`:
```
manager:~ # salt sm3.wp.suse.com inspector.export  local=True path=/tmp format=kiwi
sm3.wp.suse.com:
    The minion function caused an exception: Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1455, in _thread_return
        return_data = executor.execute()
      File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 28, in execute
        return self.func(*self.args, **self.kwargs)
      File "/usr/lib/python2.7/site-packages/salt/modules/inspector.py", line 226, in export
        raise Exception(ex)
    Exception: tostring() got an unexpected keyword argument 'pretty_print'
```

The other traceback is possible if FS type was not detected on the partition.

Fixes: https://github.com/SUSE/spacewalk/issues/8481

### Previous Behavior
Tracebacks on calling `inspector.export`

### New Behavior
Normal module behavior

### DIFF
--- a/salt/modules/inspectlib/kiwiproc.py
+++ b/salt/modules/inspectlib/kiwiproc.py
@@ -82,7 +82,7 @@ class KiwiExporter(object):
             [
                 line
                 for line in minidom.parseString(
-                    etree.tostring(root, encoding="UTF-8", pretty_print=True)
+                    etree.tostring(root, encoding="UTF-8")
                 )
                 .toprettyxml(indent="  ")
                 .split("\n")
--- a/salt/modules/inspectlib/fsdb.py
+++ b/salt/modules/inspectlib/fsdb.py
@@ -140,7 +140,7 @@ class CsvDB(object):
         return self._tables.keys()
 
     def _load_table(self, table_name):
-        with gzip.open(os.path.join(self.db_path, table_name), "rb") as table:
+        with gzip.open(os.path.join(self.db_path, table_name), "rt") as table:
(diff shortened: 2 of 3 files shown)


salt/modules/inspectlib/kiwiproc.py
```diff
@@ -82,7 +82,7 @@ def export(self, name):
             [
                 line
                 for line in minidom.parseString(
-                    etree.tostring(root, encoding="UTF-8", pretty_print=True)
```
meaksh, line 85: Do we need to drop this also here? This is 3002.2 and should be only for python3

vzhestkov: there is no pretty_print keyword parameter there and it's actually converting to pretty in the next line anyway. It was dropped by upstream and already not present in 3004

meaksh: ACK

meaksh (approved):

PR #485, opened 2022-02-07
Add missing "ansible" module functions to whitelist in Salt 3004 (bsc#1195625)

meaksh:
### What does this PR do?

This PR fixes an issue with `ansiblegate` module in Salt 3004. There is a new whitelist for the exposed commands coming from "ansiblegate" module.

Some of the latest implemented functions are missing there: `ansible.targets` and `ansible.discover_playbooks`.

Therefore, if we call any of these functions we will get:

```
2022-02-07 11:12:46,223 [salt.utils.lazy  :103 ][DEBUG   ][19401] Could not LazyLoad ansible.targets: 'ansible.targets' is not available.
```

### DIFF
--- a/salt/modules/ansiblegate.py
+++ b/salt/modules/ansiblegate.py
@@ -45,7 +45,14 @@ hosts:
 """
 DEFAULT_TIMEOUT = 1200  # seconds (20 minutes)
 
-__load__ = __non_ansible_functions__ = ["help", "list_", "call", "playbooks"][:]
+__load__ = __non_ansible_functions__ = [
+    "help",
+    "list_",
+    "call",
+    "playbooks",
+    "discover_playbooks",
+    "targets",
+][:]
 
 
 def _set_callables(modules):
--- a/salt/modules/ansiblegate.py
+++ b/salt/modules/ansiblegate.py
@@ -45,14 +45,16 @@ hosts:
 """
 DEFAULT_TIMEOUT = 1200  # seconds (20 minutes)
 
-__load__ = __non_ansible_functions__ = [
+__non_ansible_functions__ = []
+
+__load__ = __non_ansible_functions__[:] = [
     "help",
     "list_",
     "call",
     "playbooks",
     "discover_playbooks",
     "targets",
-][:]
+]
 
 
 def _set_callables(modules):


salt/modules/ansiblegate.py
```diff
@@ -45,7 +45,14 @@
...
 
-__load__ = __non_ansible_functions__ = ["help", "list_", "call", "playbooks"][:]
+__load__ = __non_ansible_functions__ = [
+    "help",
+    "list_",
+    "call",
+    "playbooks",
+    "discover_playbooks",
+    "targets",
+][:]
```
agraul, line 55: What's the point of `[:]`? With or without it, `__load__` and `__non_ansible_functions__` both point to the same object

meaksh: Hmmm, that code was actually introduced by upstream (not sure why exactly), so I simply added new values there and didn't take care of thinking on that.

Now that you mentioned that, it's true that we are actually getting the same object in both `__load__` and `__non_ansible_functions__`, which maybe it is not what was intended when using `[:]`.

In case we do `__load__ = __non_ansible_functions__[:] = ["item1", "item2"]` then we would have different objects, but not sure the reason behind this. Maybe something worth to ask upstream

meaksh: I think this could be simply a way to make both variables not to be the same instance as the static declarared list. I don't know the actual reason of this, but I see this is common in the Salt codebase.

I'll try to get some feedback in the upstream PR https://github.com/saltstack/salt/pull/60056 but I would suggest to not block this PR until getting an answer about it. The current changes on this PR are already fixing the problem we have in SUMA HEAD on the ansible features testing.

If there is any refactor on this area in the future, we can push it later if we consider it. 

Thanks for the review! :+1:

meaksh: Ok, this was clarified by upstream here https://github.com/saltstack/salt/pull/60056#issuecomment-1031457524

I'm pushing the fix for this also in this PR.

agraul (approved):
Looks correct. afaiu, `__load__` is used to let the lazy loader know about available ansible modules that should also be loaded by it.

PR #486, opened 2022-02-07
Fix "state.orchestrate_single" to not pass pillar=None (3000)

meaksh:
### What does this PR do?

Passing a pillar to state.single can results in state functions not
working when they don't accept a pillar keyword argument. One example
where this is the case is salt.wait_for_event. When no pillar is
provided, it does not need to be passed.

Upstream PR: https://github.com/saltstack/salt/pull/61093


agraul:
[e18139f](https://github.com/openSUSE/salt/pull/486/commits/e18139fc908e65dea3593d792e1dccc498285c4a) says the fix is for 3002.2, but this is the PR for 3000. What's correct?

Either way, this is only needed on the master (salt runner). IMO we don't need to push it to `3000(\.3)?`

meaksh:
@agraul since this is needed for "salt-master" in SLE12, are you fine approving this backport? Thanks in advance!

PR #489, opened 2022-02-07
Fix "state.orchestrate_single" to not pass pillar=None (3000.3)

meaksh:
### What does this PR do?

Passing a pillar to state.single can results in state functions not
working when they don't accept a pillar keyword argument. One example
where this is the case is salt.wait_for_event. When no pillar is
provided, it does not need to be passed.

Upstream PR: https://github.com/saltstack/salt/pull/61093


agraul:
IMO not needed as the fix is only for a salt runner.

meaksh:
As discussed, we are not merging this PR as the fix is not really needed on the context of `py27-compat-salt`. Closing PR.

PR #493, opened 2022-02-14
Add salt-ssh support with venv-salt-minion - 3004

vzhestkov:
Port of https://github.com/openSUSE/salt/pull/476 for `3004`

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1032,9 +1032,7 @@ class Single:
         stdout, stderr, retcode = self.shell.exec_cmd(
             "test -f {}".format(self.venv_hash_file)
         )
-        if retcode != 0:
-            return False
-        return True
+        return retcode == 0
 
     def deploy(self):
         """
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -219,7 +219,7 @@ class SSH:
 
     ROSTER_UPDATE_FLAG = "#__needs_update"
 
-    def __init__(self, opts):
+    def __init__(self, opts, context=None):
         self.__parsed_rosters = {SSH.ROSTER_UPDATE_FLAG: True}
         pull_sock = os.path.join(opts["sock_dir"], "master_event_pull.ipc")
         if os.path.exists(pull_sock) and zmq:
@@ -246,7 +246,7 @@ class SSH:
             else "glob"
         )
         self._expand_target()
-        self.roster = salt.roster.Roster(self.opts, self.opts.get("roster", "flat"))
+        self.roster = salt.roster.Roster(self.opts, self.opts.get("roster", "flat"), context=context)
         self.targets = self.roster.targets(self.opts["tgt"], self.tgt_type)
         if not self.targets:
             self._update_targets()
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -719,6 +719,11 @@ class SSH:
                             )
                             ret = {"id": host, "ret": error}
                             log.error(error)
+                            log.error(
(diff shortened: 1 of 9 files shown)


salt/client/ssh/__init__.py
```diff
@@ -580,7 +588,7 @@ def handle_ssh(self, mine=False):
         """
         que = multiprocessing.Queue()
         running = {}
-        target_iter = self.targets.__iter__()
+        targets_queue = list(self.targets.keys())
```
agraul, line 598: Why is it needed to create a list here?

salt/client/ssh/__init__.py
```diff
@@ -589,11 +597,40 @@ def handle_ssh(self, mine=False):
                 log.error("No matching targets found in roster.")
                 break
             if len(running) < self.opts.get("ssh_max_procs", 25) and not init:
-                try:
-                    host = next(target_iter)
-                except StopIteration:
+                if targets_queue:
+                    host = targets_queue.pop(0)
```
agraul, line 608: How large is `targets_queue`? `list.pop(0)` copies all pointers of the remaining elements, is that a problem here?

agraul (commented):
I have a few questions before about the PR, it would be great if we could clarify them before I take another look:

1. Is `LOG_LOCK` needed with the changes to salt loader done between 3002.2 and 3004?
2. Do we already have an upstream PR for the `ssh_pre_flight_args` ?
3. Could you squash the commits please? I think there are enough unrelated changes to warrant more than a single commit, but I think 17 are a bit much

salt/client/ssh/__init__.py
vzhestkov: It's used to put the blocked minion at the end of the list to retry it after:
https://github.com/openSUSE/salt/blob/03b0e9bd05cfeb508d78c2ed95a2710310af5cdb/salt/client/ssh/__init__.py#L632

salt/client/ssh/__init__.py
vzhestkov: Maximum size of the list is the number of salt-ssh systems to run the events on. In almost 99% it's just 1 element. In Uyuni there is a possibility to run salt-ssh event for multiple minions in one event is to use `Remote command` with wildcard, so the maximum number of the list in this case is the number of registered salt-ssh systems on the Server.

agraul: Then I think this is fine, otherwise we could swap it out for a https://docs.python.org/3/library/collections.html#collections.deque, which actually is a linked list and not an array of pointers.

vzhestkov:
> Is LOG_LOCK needed with the changes to salt loader done between 3002.2 and 3004?

Yes, it is. Otherwise it gets deadlock in case if logging set to DEBUG or TRACE

> Do we already have an upstream PR for the ssh_pre_flight_args ?

Not yet.


I'll squash it, I used distinct commits to revert some of them easier.

salt/client/ssh/__init__.py
vzhestkov: @agraul could you please check https://github.com/openSUSE/salt/pull/493/commits/d2564f1954ad6ab253575390443c704e638ea51a

In stress test looks fine.

meaksh (approved):

agraul:
> > Do we already have an upstream PR for the ssh_pre_flight_args ?
> 
> Not yet.
> 
> I'll squash it, I used distinct commits to revert some of them easier.

Can we take care of these two action items before we merge the PR?

vzhestkov:
I will squash all the commits on merging.
Upstream PR for `ssh_pre_flight_args` is https://github.com/saltstack/salt/pull/61715
Will create distinct PR for other features required for `venv-salt-minion` handling later. @agraul is it fine to do it this way?

PR #506, opened 2022-04-11
Fix regression with depending on psutil (bsc#1197533) - 3004

vzhestkov:
### What does this PR do?

With Salt SSH simultaneous sessions handling a regression was introduced, causing dependency for `psutil` python module on Salt SSH client.
The dependency is indirect and really hard to notice: `salt.modules.saltutils` requires `salt.client.ssh` and it requires `psutil`, but `saltutils` is used on Salt SSH system directly and the traceback on loading this module is masked and only wisible in DEBUG logging level on Salt SSH client.

Fixes: https://github.com/SUSE/spacewalk/issues/17375

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -12,7 +12,6 @@ import hashlib
 import logging
 import multiprocessing
 import os
-import psutil
 import queue
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


meaksh (approved):
Thanks :+1:

PR #509, opened 2022-04-21
Use Salt Bundle in dockermod - 3004

vzhestkov:
### What does this PR do?

The PR implements using the Salt Bundle for `dockermod.sls` calls inside the container in case if it called on the host running the Salt Bundle. All the calls inside the container are performed with the Bundle shipped inside with the tarball.

Upstream PR: https://github.com/saltstack/salt/pull/62109
Fixes: https://github.com/SUSE/spacewalk/issues/17507

### Previous Behavior
Fails in case if there is no reqired python version present inside the container.

### New Behavior
One limitation still present: as it uses the Bundle from the host the container need to be the same OS as on the host. So it's not possible to build SLE12 images on SLE15 hosts.

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/master/topics/development/contributing.html) for best practices.

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -201,14 +201,19 @@ import copy
 import fnmatch
 import functools
 import gzip
+import hashlib
 import json
 import logging
 import os
+import pathlib
 import pipes
 import re
 import shutil
 import string
 import subprocess
+import sys
+import tarfile
+import tempfile
 import time
 import uuid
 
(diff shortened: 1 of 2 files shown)


salt/modules/dockermod.py
```diff
@@ -6717,47 +6827,68 @@ def call(name, function, *args, **kwargs):
...
+        run_all(
+            name,
+            subprocess.list2cmdline(
+                ["tar", "zxf", dest_venv_tar, "-C", venv_dest_path]
+            ),
+        )
+        run_all(name, subprocess.list2cmdline(["rm", "-f", dest_venv_tar]))
+        container_python_bin = venv_python_bin
+        thin_dest_path = os.path.join(venv_dest_path, venv_name)
+        thin_salt_call = os.path.join(thin_dest_path, "bin", "salt-call")
```
meaksh, line 6847: Hmmm, small nitpick here: could you rename these variable to something like `container_dest_path` and `container_salt_call` or something like that?

I feel a bit confusing that we use `thin` a part of the variable names here since we are actually going into the alternative approach to the salt thin (venv approach).

salt/modules/dockermod.py
```diff
@@ -6717,47 +6827,68 @@ def call(name, function, *args, **kwargs):
...
+        if ret["retcode"] != 0:
+            return {"result": False, "comment": ret["stderr"]}
+        thin_salt_call = os.path.join(thin_dest_path, "salt-call")
 
     try:
         salt_argv = (
             [
                 container_python_bin,
-                os.path.join(thin_dest_path, "salt-call"),
+                thin_salt_call,
```
meaksh, line 6891: Same reasoning as above. I would choose `container_salt_call` here instead of `thin_salt_call` for better consistency of variable names.

nodeg:
Any news on this?

meaksh (approved):
Let's give this a try, thanks :+1:

PR #518, opened 2022-05-17
Make sure SaltCacheLoader use correct fileclient

witekest:
Backported from https://github.com/saltstack/salt/pull/61895

Fixes: https://github.com/SUSE/spacewalk/issues/17708


witekest:
According to saltstack/salt#60003 Salt 3002 is not affected.

PR #519, opened 2022-05-18
Make sure SaltCacheLoader use correct fileclient

witekest:
Backported from https://github.com/saltstack/salt/pull/61895

Fixes: https://github.com/SUSE/spacewalk/issues/17708

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -4061,6 +4061,9 @@ class BaseHighState:
             )
         else:
             try:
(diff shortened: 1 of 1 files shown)


meaksh:
It seems this is only affecting Salt 3004, the Salt 3002.2 package should be fine according to https://github.com/saltstack/salt/issues/60003

Let's wait until having test results for this PR.

PR #529, opened 2022-05-25
[WIP] Make the pass renderer more configurable

darix:
### What does this PR do?
Make the pass renderer more configurable

1. allow us to make the pass renderer fail hard during pillar rendering
   For this we add a master config variable pass_renderer_fail_hard

2. allow to have prefix for variables that should be passed through pass
   For this we add a master config variable pass_variable_prefix

   This would lower the number of false positives triggered in the hard
   fail case.


salt/renderers/pass.py
```diff
@@ -84,9 +94,13 @@ def _fetch_secret(pass_path):
...
     if proc.returncode or not pass_data:
-        log.warning("Could not fetch secret: %s %s", pass_data, pass_error)
+        if "pass_renderer_fail_hard" in __opts__ and __opts__["pass_renderer_fail_hard"]:
+            raise SaltRenderError("Unable to retrieve secret '{}' from the password store: {}".format(pass_path, stderr))
+        else:
+            log.warning("Could not fetch secret: %s %s", pass_data, pass_error)
         pass_data = pass_path
-    return pass_data.strip()
+    # only strip trailing whitespaces
+    return pass_data.rstrip('\r\n')
```
meaksh, line 103: According to the comment, you want to strip only trailing whitespaces, then shouldn't we use simply `rstrip()` here?

Passing `\r\n` won't strip trailing whitespaces in case our string is like `test   `.

darix: yeah we only want to strip newlines so the password can end in a ` `

meaksh: Ok :+1: let's fix the comment then to avoid confussion

meaksh:
@darix I guess we can close this PR in favor of https://github.com/openSUSE/salt/pull/532, correct?

darix:
yes. #532 was done by Daniel and me.

PR #539, opened 2022-06-24
add support for gpgautoimport

mcalmer:
### What does this PR do?

- add support for gpgautoimport to refresh_db in the zypperpkg module
- call refresh_db function from mod_repo
- call refresh_db with kwargs where possible

Upstream PR https://github.com/saltstack/salt/pull/62209


SchoolGuy (approved):
Since this is just a backport, this looks good. The upstream PR is reviewed and seems to look also good.

PR #559, opened 2022-09-29
 Include stdout in error message for zypperpkg

ycedres:
### What does this PR do?
For the zypperpkg module, improves error information to the user if zypper returns with an error code.

Backported from https://github.com/saltstack/salt/pull/62750
Fixes: https://github.com/SUSE/spacewalk/issues/17746


vzhestkov (approved):
Looks good, but seems this PR is a backport not of https://github.com/saltstack/salt/pull/62750 only but of https://github.com/saltstack/salt/pull/62346 also. Maybe it make some sense to mention it in the spec.

PR #562, opened 2022-10-06
[skip] adjust github workflows after branch renaming

meaksh:
### What does this PR do?

This PR fixes the GitHub actions for `openSUSE/release/3004` after so they run after branch renaming.

### Commits signed with GPG?
Yes

### DIFF
--- a/.github/workflows/integration-opensuse-leap.yml
+++ b/.github/workflows/integration-opensuse-leap.yml
@@ -3,10 +3,10 @@ name: integration-opensuse-leap
 on:
   push:
     branches:
-      - openSUSE-3004
+      - openSUSE/release/3004
   pull_request:
     branches:
-      - openSUSE-3004
+      - openSUSE/release/3004
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-opensuse-leap.yml
```diff
@@ -1,14 +1,5 @@
 name: integration-opensuse-leap
-
-on:
-  push:
-    branches:
-      - openSUSE-3004
```
agraul, line 6: Why is it "okay" to remove naming these branches? Doesn't this cause double-executions on PRs from this repository?

.github/workflows/integration-opensuse-leap.yml
```diff
@@ -1,14 +1,5 @@
...
-  pull_request:
-    branches:
-      - openSUSE-3004
-  workflow_dispatch:
```
agraul, line 10: IMO it is nice to have a way to manually trigger the tests, but I have to admit I haven't used it in a long time. Does it hurt keeping this?

agraul (changes requested):
I'm not sure this is the best way to fix our issue, especially when we push to branches in this repository (opposed to forks)

.github/workflows/integration-opensuse-leap.yml
meaksh: Agree. I does not hurt keeping it :+1:

agraul (approved):

PR #564, opened 2022-10-06
3000: Prevent double executions of GH actions

meaksh:
### What does this PR do?
This PR fixes the GH actions for the `openSUSE/release/3000` branch to avoid double executions of tests on the PRs.

### Commits signed with GPG?
Yes

### DIFF
--- a/.github/workflows/integration-opensuse-leap-15-1.yml
+++ b/.github/workflows/integration-opensuse-leap-15-1.yml
@@ -1,5 +1,13 @@
 name: integration-opensuse-leap-15-1
-on: [push, pull_request]
+on:
+  push:
+    branches:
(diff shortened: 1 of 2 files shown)


agraul:
I think this is not needed. Maybe I missed it, but I didn't see any duplicate runs for this PR in the [overview](https://github.com/openSUSE/salt/actions)

meaksh:
Hmm, I thought I saw this happening also for 3000 branch but it seems I was wrong.

I wonder why we do not have duplicated executions here and we did at 3004 branch when the branch name was not set. I may be missing something.

meaksh:
I'm closing this now as it seems it not really needed.

PR #586, opened 2023-06-15
Mark Salt 3006 as released

nodeg:
### What does this PR do?

Sets Salt 3006 as released.

Fixes:

### Previous Behavior

```bash
salt '*' salt_version.equal 'Sulfur'
```
reports False although Salt 3006 is used.

### New Behavior

Without this, commands like

```bash
salt '*' salt_version.equal 'Sulfur'
```

returns True if Salt 3006 is used.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


nodeg:
Not sure why the tests broke. Any idea @meaksh, @agraul?

agraul:
> Not sure why the tests broke. Any idea @meaksh, @agraul?

No idea why the are broken with your PR. I think this was the reason @meaksh reverted your change when he prepared Salt 3006.0. We probably all agree that we need to find the real fix instead of keeping 3006.0 marked as unreleased., right?

nodeg:
I agree.

meaksh:
In the meanwhile, I was investigating a bit more on this and realized about the `_version.py`. This, should set a fixed version for Salt.

This is what Salt upstream uses in the GH actions to set a fixed Salt version. We also used this mechanism in the past, until Salt 3004, but maybe we still need to make use of it.

I'm just currently doing some tests with a fixed version provided via `_version.py` in a test package: https://build.opensuse.org/package/show/home:PSuarezHernandez:branches:systemsmanagement:saltstack:products:testing/salt

meaksh:
Ok, I think I've to understand what is going on here.

There are 2 different problems here:

1. There is an upstream bug that makes `salt_version.equal` and other methods to not detect the codename properly. If you check `salt_version.equal Sulfur` in upstream Salt 3006.1 (installed via pip), it also reports `False`. I managed to fix it, will produce upstream PR and push commits here as well.

2. We set the fixed version when building the Salt package (running `setup.py --with-salt-version=XXXX`: https://github.com/openSUSE/salt-packaging/blob/release/3006.0/salt/salt.spec#L723. This generates the mentioned `_version.txt` that provides a fixed version for the Salt build. The problem is that, in GH actions we use Salt Toaster passing a custom `SALT_REPO=` flag, therefore we do not use a built package (containing the `_version.txt`) but instead we run tests directly on top of the GH repo (without fixing the version). This makes GH actions tests to mess with the version. So, in order to fix this, we would probably need to make either GH actions or Salt Toaster (probably toaster) to fix the Salt version.

meaksh:
FYI I've created these PRs:

- https://github.com/openSUSE/salt-toaster/pull/99
- https://github.com/saltstack/salt/pull/64554

I will backport some commits from upstream PR to this PR in order to solve the version mess here, and make tests to pass.

meaksh:
This is ready to review.

As mentioned this is partially backporting https://github.com/saltstack/salt/pull/64554 to fix the problem with the codename and version detection.

agraul (approved):

nodeg:
Thank you for improving and taking this over, @meaksh!

PR #594, opened 2023-08-09
Implement the calling for batch async from the salt CLI

vzhestkov:
### What does this PR do?

There was no way to call batch async from salt CLI, in case of specifying `batch` and `async` in the same call the event was processing with `batch`.
This PR implements the call for batch async in case if `batch` and `async` both specified.

### Previous Behavior
In case of specifying `batch` and `async` in the call, `batch` was used and `async` was ignored.

### New Behavior
Works as expected

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/cli/salt.py
+++ b/salt/cli/salt.py
@@ -47,7 +47,12 @@ class SaltCMD(salt.utils.parsers.SaltCMDOptionParser):
             self.exit(2, "{}\n".format(exc))
             return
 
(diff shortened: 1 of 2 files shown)


meaksh:
This is weird... but apparently the changes on this PR are making the some unrelated unit tests to crash.

Could you have a look? Thanks!

PR #599, opened 2023-08-31
Fixed gitfs cachedir_basename to avoid hash collisions (bsc#1193948, bsc#1214797, CVE-2023-20898)

meaksh:
### What does this PR do?

This PR backports the fixes for CVE-2023-20898 from upstream PR https://github.com/saltstack/salt/pull/64959 to our `openSUSE/release/3006.0` branch.

Tracks: https://github.com/SUSE/spacewalk/issues/16594

- [ ] Docs
- [x] Changelog - https://docs.saltproject.io/en/master/topics/development/changelog.html
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -3,6 +3,7 @@ Classes which provide the shared base for GitFS, git_pillar, and winrepo
 """
 
 
+import base64
 import contextlib
 import copy
 import errno
@@ -11,10 +12,12 @@ import glob
 import hashlib
 import io
 import logging
+import multiprocessing
 import os
 import shlex
 import shutil
 import stat
+import string
 import subprocess
 import time
 import weakref
@@ -22,6 +25,7 @@ from datetime import datetime
 
 import salt.ext.tornado.ioloop
 import salt.fileserver
+import salt.syspaths
 import salt.utils.configparser
 import salt.utils.data
 import salt.utils.files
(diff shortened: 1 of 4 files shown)


salt/utils/gitfs.py
```diff
@@ -838,6 +895,20 @@ def _lock(self, lock_type="update", failhard=False):
...
+        finally:
+            self.__class__._master_lock.release()
+
+    def __lock(self, lock_type="update", failhard=False):
+        """
+        Place a lock file if (and only if) it does not already exist.
+        Without MultiProcessing locks.
+        """
         try:
             fh_ = os.open(
```
vzhestkov, line 913: I'm a bit confused why `salt.utils.files.flopen` was not used there to prevent any possible race conditions with the file creation.

vzhestkov (approved):
Just one nitpicky comment in the code, but not really related to this change, but to the approach used to create lock files in general.
Other than that it looks good to me.

salt/utils/gitfs.py
meaksh: Thanks for the review! And good point.

Hmmm, I have also no idea. Maybe this is something to have in mind and check in case we still detect more race conditions around this area.

For now, in the context of the present submission and CVE fix, I would say to proceed with the current changes. Later we can evaluate and test if we should also change that.

PR #606, opened 2023-10-09
Remove usage of "mock" in favor of "unittest.mock"

meaksh:
### What does this PR do?

This PR simply removes the usage of `mock` in favor of `unittest.mock` (standard library), as after Salt dropping support for Python 3.6, there is no need to keep using `mock` library at the time of testing.

Additionally, it removes the import of  `__version__`  (an unused `__mock_version`), as this attribute was dropped and not available in Python 3.9+ (https://github.com/python/cpython/pull/17977)

---

This PR backports https://github.com/saltstack/salt/pull/65364 to `openSUSE/release/3006.0` branch.

### Commits signed with GPG?
Yes

### DIFF
--- a/requirements/pytest.txt
+++ b/requirements/pytest.txt
@@ -1,4 +1,3 @@
-mock >= 3.0.0
 # PyTest
 pytest >= 7.0.1; python_version <= "3.6"
 pytest >= 7.2.0; python_version > "3.6"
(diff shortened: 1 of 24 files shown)


meaksh:
We cannot yet replace `mock` in favor of `unittest.mock` yet, as tests are aligned with latest `mock` version functionality (not available in Python 3.6 unittest.mock`.

Closing this PR without merging for now.

PR #611, opened 2023-10-31
Dereference symlinks to set proper __cli opt (bsc#1215963)

meaksh:
This PR backports https://github.com/saltstack/salt/pull/65435 to `openSUSE/release/3006.0`.

---

### What does this PR do?
In case of creating symlinks to `salt-call`, the behaviour of some modules could be different than expected as `__cli` contains the name of the symlink instead of `salt-call` itself. This change will dereference the symlinks to set `__cli` opt to the proper value.

### Previous Behavior
Different behaviour on calling modules like `state`, `event` and `git` if the call is performed with symlink to `salt-call` instead of `salt-call` itself.

### New Behavior
The same behaviour for `salt-call` and any name of symlink pointint to it as expected.

### DIFF
--- a/changelog/65435.fixed.md
+++ b/changelog/65435.fixed.md
@@ -1,0 +1,1 @@
+Dereference symlinks to set proper __cli opt
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -3747,7 +3747,9 @@ def apply_minion_config(
             )
             opts["fileserver_backend"][idx] = new_val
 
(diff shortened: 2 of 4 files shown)


vzhestkov:
@meaksh not sure if I understood the relation between this change and usage of mock. Is there really such relation or the description is from the other change?

meaksh:
> @meaksh not sure if I understood the relation between this change and usage of mock. Is there really such relation or the description is from the other change?

Oops, I made a mistake when copying the PR description :cry: I'll update it. :)

vzhestkov (approved):
Thank you for taking care of it :+1:

PR #612, opened 2023-11-14
Use unittest.mock for newer Python versions

ycedres:
### What does this PR do?

This PR causes the test suite to use unittest.mock instead of mock for supported Python versions, so the usage of standard library components is favoured.

To achieve this, the __version__ variable is removed as it was dropped from Python 3.9 and it's almost not used.

The results of the test suite do not change.

Upstream PR:
https://github.com/saltstack/salt/pull/65644

Related: https://github.com/SUSE/spacewalk/issues/22793

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/support/mock.py
+++ b/tests/support/mock.py
@@ -18,36 +18,32 @@ import copy
 import errno
 import fnmatch
 import sys
+import importlib
 
-# By these days, we should blowup if mock is not available
-import mock  # pylint: disable=blacklisted-external-import
+current_version = (sys.version_info.major, sys.version_info.minor)
 
-# pylint: disable=no-name-in-module,no-member
-from mock import (
-    ANY,
-    DEFAULT,
-    FILTER_DIR,
-    MagicMock,
-    Mock,
-    NonCallableMagicMock,
-    NonCallableMock,
-    PropertyMock,
-    __version__,
-    call,
-    create_autospec,
-    patch,
-    sentinel,
-)
+# Prefer unittest.mock for Python versions that are sufficient
+if current_version >= (3,8):
+    mock = importlib.import_module('unittest.mock')
+else:
+    mock = importlib.import_module('mock')
+
+ANY = mock.ANY
+DEFAULT = mock.DEFAULT
+FILTER_DIR = mock.FILTER_DIR
+MagicMock = mock.MagicMock
+Mock = mock.Mock
+NonCallableMagicMock = mock.NonCallableMagicMock
+NonCallableMock = mock.NonCallableMock
+PropertyMock = mock.PropertyMock
+call = mock.call
+create_autospec = mock.create_autospec
+patch = mock.patch
+sentinel = mock.sentinel
 
 import salt.utils.stringutils
 
 # pylint: disable=no-name-in-module,no-member
-
-
-__mock_version = tuple(
-    int(part) for part in mock.__version__.split(".") if part.isdigit()
-)  # pylint: disable=no-member
-
 
 class MockFH:
     def __init__(self, filename, read_data, *args, **kwargs):
--- a/requirements/pytest.txt
+++ b/requirements/pytest.txt
@@ -1,4 +1,4 @@
-mock >= 3.0.0
+mock >= 3.0.0; python_version < '3.8'
 # PyTest
 pytest >= 7.0.1; python_version <= "3.6"
 pytest >= 7.2.0; python_version > "3.6"
--- a/tests/pytests/unit/cloud/clouds/test_dimensiondata.py
+++ b/tests/pytests/unit/cloud/clouds/test_dimensiondata.py
@@ -11,7 +11,6 @@ from salt.cloud.clouds import dimensiondata
 from salt.exceptions import SaltCloudSystemExit
 from salt.utils.versions import Version
 from tests.support.mock import MagicMock
-from tests.support.mock import __version__ as mock_version
 from tests.support.mock import patch
 
 try:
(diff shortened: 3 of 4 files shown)


tests/support/mock.py
```diff
@@ -18,26 +18,32 @@
...
-    create_autospec,
-    patch,
-    sentinel,
-)
+import importlib
+
+current_version = (sys.version_info.major, sys.version_info.minor)
+
+# Use python-mock for older Python versions that are not longer supported
+# in the standard library.
```
agraul, line 26: This comment is slightly misleading. `python-mock` is a rolling backport of CPython's main branch and can contain features that have not been backported to supported Python versions. Right now, `python-mock-5.1.0` includes `ThreadingMock`. That's a new class that was only added a few months ago and is not in any supported Python release:
```
>>> from unittest.mock import ThreadingMock
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'ThreadingMock' from 'unittest.mock' (/usr/lib64/python3.12/unittest/mock.py)
>>> from mock import ThreadingMock
>>> 
```

agraul (commented):
I think the approach makes sense since we can only use the version number to conditionally set a dependency to `python3-mock` in the spec file.

Note: You have a "Fixes: GH issue" in the PR descripition. That issue won't be fixed by this PR, it's about pushing packages to openSUSE:Factory. Don't forget to re-open once you merge this PR!

tests/support/mock.py
```diff
@@ -18,26 +18,32 @@
...
+if current_version < (3,9):
+    mock = importlib.import_module('mock')
+else:
+    mock = importlib.import_module('unittest.mock')
+    # The __version__ variable is no longer available from Python 3.9
+    mock.__version__ = sys.version
```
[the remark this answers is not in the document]
ycedres: > Up to 3.9 (excluded) this was always `'1.0'`. With this patch it's something like `'3.11.5 (main, Aug 31 2023, 07:57:41) [GCC]'`. Further down that is then parsed to `(3,11)`.

I'm sorry I didn't understand this. What was `1.0` before 3.9?

> 
> For `mock`, `mock.__version__` is something like `'5.1.0'`, parsed to `(5,1,0)`.
> 
> I haven't checked where `__mock_version` is used, is a two-item tuple okay?

 I believe it's never used at this moment.  It [was](https://github.com/openSUSE/salt/blob/4d4901324709e49ad81f3903e86904f3855aec5e/tests/support/mock.py#L50) used to require mock >= 2.0 for the oldest Python 2 versions.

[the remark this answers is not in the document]
agraul: > I'm sorry I didn't understand this. What was 1.0 before 3.9?
`unittest.mock.__version__`

tests/support/mock.py
ycedres: So if upstream Salt uses (now or in the future) features in `python-mock` not yet forward-ported to `unittest.mock` in stdlib then we would have a problem anyway running the test suite in Tumbleweed, is that correct?

ycedres:
BTW @agraul about the topic of creating this PR upstream, my concern is that if Salt 3006.4 (at this moment) requires at least Python 3.7 and this Python version is no longer supported then `unittest.mock` could not be used with 3.7. So you will need this upstream (pseudocode): 

```
if python-version < 3.8:
   import mock
else:
   import unittest.mock
```
   
However, requiring at least Python 3.9 because it's the lower version available in TW is our requirement but not theirs. That's why I'm not sure if this is good for upstream, how do you see it?

tests/support/mock.py
agraul: Yes, if Salt uses features in `python-mock` that have not been backported from CPython@main to CPython@3.X where 3.X is all versions in TW, we have a problem running the testsuite. This is not very likely to cause big issues, if you look at the history of `unittest.mock`, new mock classes are not added often. And when they are, they cover very specific cases and Salt first needs to adopt them. Let's ignore that for now.

agraul:
> BTW @agraul about the topic of creating this PR upstream, my concern is that if Salt 3006.4 (at this moment) requires at least Python 3.7 and this Python version is no longer supported then `unittest.mock` could not be used with 3.7. So you will need this upstream (pseudocode):
> 
> ```
> if python-version < 3.8:
>    import mock
> else:
>    import unittest.mock
> ```
> 
> However, requiring at least Python 3.9 because it's the lower version available in TW is our requirement but not theirs. That's why I'm not sure if this is good for upstream, how do you see it?

I don't really understand the problem. Why do we have to set it to `python-version < 3.9` for TW? Our goal is simply to use `unittest.mock` if it is good enough and fall back to `python-mock` when it isn't. I would compare `python-version` to the lowest Python version that contains a `unittest.mock` library that's good enough.

ycedres:
I run the upstream Salt test suiste (using Nox - Python 3.8) and I got the same result with and without the changes in this PR. Same conclusion with running the `openSUSE/salt-test` test suite in SLE with Python 3.6. In my view that would make 3.8 fit that definition of <em>good enough</em> since 3.8 is the older supported version.

meaksh (approved):
Looks good to me. Thanks.

Let's wait for Alex's feedback as he has been actively involved in this PR discussion.

agraul:
Do we already have an upstream PR?

ycedres:
> Do we already have an upstream PR?

There is one now: https://github.com/saltstack/salt/pull/65644

Maybe @meaksh wants to close the old one related to this: https://github.com/saltstack/salt/pull/65364

meaksh:
I just closed my upstream PR in favor of this new one.

BTW @ycedres shouldn't we remove the dependency to `mock` in the requirements files for Python versions >= 3.8 ?

agraul (approved):
LGTM!

PR #614, opened 2023-11-15
Revert "Make sure configured user is properly set by Salt" (bsc#1216284)

meaksh:
This reverts commit 5ea4add5c8e2bed50b9825edfff7565e5f6124f3 which turned out to be problematic, as it caused some regressions (bsc#1216284), while not being really needed to fix the initial issue that motivated that patch (bsc#1210994).

The patch this PR is reverting is not really needed to fix (bsc#1210994), as that bug was actually fixed by https://github.com/openSUSE/salt/commit/40a57afc65e71835127a437248ed655404cff0e8. 

Introducing the conflictive patch was just to align with upstream PR https://github.com/saltstack/salt/pull/64510, after it was reviewed (which contains the initial working fix + extra changes after the review)

The reason to revert this patch instead of fixing upstream PR is that, it seems we cannot really change the "user" that earlier in the stack as proposed in upstream PR, because this causes problems for "salt-master", while creating user keys (as "chmod" cannot be executed properly if not running as root). So the extra changes done in the upstream PR don't really make sense.

In a nutshell, "salt-master" needs to run as "root" during startup, and then it will switch to the configured user at some point during the startup. We cannot set "salt-master" systemd unit with `User=salt` and neither switching to the configured user  at the begining of the initialization stack, as this is too early and will cause problems.

I will adapt upstream PR to refllect this findings.

Tracks: https://github.com/SUSE/spacewalk/issues/22827

### DIFF
--- a/pkg/common/salt-master.service
+++ b/pkg/common/salt-master.service
@@ -8,7 +8,6 @@ LimitNOFILE=100000
 Type=notify
 NotifyAccess=all
 ExecStart=/usr/bin/salt-master
-User=salt
 
 [Install]
 WantedBy=multi-user.target
--- a/pkg/old/deb/salt-master.service
+++ b/pkg/old/deb/salt-master.service
@@ -7,7 +7,6 @@ LimitNOFILE=16384
 Type=notify
 NotifyAccess=all
 ExecStart=/usr/bin/salt-master
-User=salt
 
 [Install]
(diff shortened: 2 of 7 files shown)


vzhestkov (approved):
Thanks for taking care of it :+1: Looks good for me.

PR #616, opened 2024-01-03
Add test workflows based on salt-test

agraul:
### What does this PR do?
Replace the salt-toaster based unit and integration test workflows with salt-test based ones. While adding these workflows, I came across a number of tests I had to fix. Where possible, I cherry-picked upstream fixes.

The hostname is set to `salt-test-container`, which is used to conditionally disable tests that are problematic in GH Actions. If anyone has a better idea for either disabling these tests or know how to run them in GH Actions, I'm all for it.

Needs updates to https://build.opensuse.org/package/show/systemsmanagement:saltstack:products:testing:containers:leap15.5/salt-testsuite-ci, see [link diff](https://build.opensuse.org/package/rdiff/home:agraul:branches:systemsmanagement:saltstack:products:testing:containers:leap15.5/salt-testsuite-ci?opackage=salt-testsuite-ci&oproject=systemsmanagement%3Asaltstack%3Aproducts%3Atesting%3Acontainers%3Aleap15.5)


tests/integration/modules/test_cmdmod.py
```diff
@@ -1,634 +0,0 @@
...
-
-import salt.utils.path
-import salt.utils.platform
-import salt.utils.user
-from tests.support.case import ModuleCase
-from tests.support.helpers import SKIP_INITIAL_PHOTONOS_FAILURES, dedent
-from tests.support.runtests import RUNTIME_VARS
-
-AVAILABLE_PYTHON_EXECUTABLE = salt.utils.path.which_bin(
-    ["python", "python2", "python2.6", "python2.7"]
```
vzhestkov, line 17: Not sure if I understood why the whole test file was dropped, but I could assume due to missing python executable. Or am I wrong?

vzhestkov (approved):
In general looks good to me, just one comment about removed `test_cmdmod.py`.

tests/integration/modules/test_cmdmod.py
meaksh: These tests were actually just migrated and moved to this other file: https://github.com/openSUSE/salt/pull/616/files#diff-7f58f80c7428a46156300ebfac3a3ee8d98811d3e8e69fd26a51239c2dc6ee29

(Backported from: https://github.com/saltstack/salt/pull/65679)

PR #618, opened 2024-01-11
Allow kwargs for fileserver roots update (bsc#1218482)

ycedres:
### What does this PR do?

When an orchestration state is applied the user responsible for the runner is passed to the `update` function as the `__pub_user` keyword argument. Therefore, the `update` function must be aware of this or the orchestration state application will fail.

Fixes: https://github.com/SUSE/spacewalk/issues/23324
Upstream PR: https://github.com/saltstack/salt/pull/65819

### Previous Behavior
The application of this example orchestration state fails

```
my_orch:
  salt.runner:
    - name: fileserver.update
```
with the following error:
```
 Passed invalid arguments: update() got an unexpected keyword argument '__pub_user'
```
### New Behavior
Now the orchestrating state is applied correctly.

Additionally, it fixes wrong asserts on fileserver.update tests, as it was producing false positives when ret["return"] is a string (i.a. when errors are produced) instead of True value.

- [ ] Docs
- [X] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/fileserver/roots.py
+++ b/salt/fileserver/roots.py
@@ -138,16 +138,10 @@ def serve_file(load, fnd):
     return ret
 
 
-def update(**kwargs):
+def update():
     """
     When we are asked to update (regular interval) lets reap the cache
     """
-    # __pub_user responsible for the runner can be passed but we don't do anything with it
-    if "__pub_user" in kwargs:
-        del kwargs["__pub_user"]
-    if kwargs:
-        raise ValueError("Unexpected keyword arguments received: %s" % kwargs)
-
     try:
         salt.fileserver.reap_fileserver_cache_dir(
             os.path.join(__opts__["cachedir"], "roots", "hash"), find_file
--- a/salt/fileserver/roots.py
+++ b/salt/fileserver/roots.py
@@ -138,10 +138,16 @@ def serve_file(load, fnd):
     return ret
 
 
-def update():
+def update(**kwargs):
     """
     When we are asked to update (regular interval) lets reap the cache
     """
+    # __pub_user responsible for the runner can be passed but we don't do anything with it
+    if "__pub_user" in kwargs:
+        del kwargs["__pub_user"]
+    if kwargs:
+        raise ValueError("Unexpected keyword arguments received: %s" % kwargs)
+
     try:
         salt.fileserver.reap_fileserver_cache_dir(
             os.path.join(__opts__["cachedir"], "roots", "hash"), find_file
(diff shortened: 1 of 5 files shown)


salt/fileserver/roots.py
```diff
@@ -138,10 +138,16 @@ def serve_file(load, fnd):
...
-def update():
+def update(**kwargs):
     """
     When we are asked to update (regular interval) lets reap the cache
     """
+    # __pub_user responsible for the runner can be passed but we don't do anything with it
+    if "__pub_user" in kwargs:
+        del kwargs["__pub_user"]
+    if kwargs:
+        raise ValueError("Unexpected keyword arguments received")
```
agraul, line 149: It would be helpful to include `kwargs` in the exception. In the future, if something calls `update()`, it will be hard to `del` the argument from `kwargs` without knowing anything about it.

salt/fileserver/roots.py
```diff
@@ -138,10 +138,16 @@ def serve_file(load, fnd):
...
 
 
-def update():
+def update(**kwargs):
     """
     When we are asked to update (regular interval) lets reap the cache
     """
+    # __pub_user responsible for the runner can be passed but we don't do anything with it
+    if "__pub_user" in kwargs:
+        del kwargs["__pub_user"]
```
agraul, line 147: Does it make sense to manually maintain a list of args that a caller passes just to delete them?

agraul (changes requested):
Can you please add the link to the upstream PR to the description?

I left a few comments inline.

salt/fileserver/roots.py
ycedres: The issue that caused the bug was the propagation of the user responsible for the runner (see https://github.com/saltstack/salt/pull/64038) and this user is passed to the fileserver backend [here](https://github.com/openSUSE/salt/blob/107de57586f0b0f784771543b942dfb6bb70453a/salt/runners/fileserver.py#L353). So I assumed that in the future other keyword arguments may reach this point even though they are not useful. 

I wanted to have at least some control of what arguments are being passed so the user doesn't just add an argument expecting some behaviour with no effect. Of course there might be a better solution :)

salt/fileserver/roots.py
```diff
@@ -138,10 +138,16 @@ def serve_file(load, fnd):
...
+    if kwargs:
+        raise ValueError("Unexpected keyword arguments received: %s" % kwargs)
```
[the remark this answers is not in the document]
ycedres: The approach changed with the new commits by @meaksh

vzhestkov (approved):
Looks good to me :+1: 
Thanks for taking care of the tests :+1:

PR #619, opened 2024-01-17
Update __pillar__ during pillar_refresh

m-czernek:
This PR backports https://github.com/saltstack/salt/pull/65092 to `openSUSE/release/3006.0`.

### What does this PR do?

We need to make sure that we update pillar during a pillar refresh to ensure that process_beacons has the updated beacons loaded from pillar.

### Previous Behavior
Previously, users would have to restart `venv-salt-minion` process to get the updated beacons.

### New Behavior
The updated beacons are available to the  `venv-salt-minion` without restarting it.

### DIFF
--- a/changelog/63583.fixed.md
+++ b/changelog/63583.fixed.md
@@ -1,0 +1,1 @@
+Need to make sure we update __pillar__ during a pillar refresh to ensure that process_beacons has the updated beacons loaded from pillar.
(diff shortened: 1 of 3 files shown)


agraul (approved):

salt/modules/aptpkg.py
agraul, line 1: I think we should have this commit in a separate PR. Each PR will be squashed into a single commit and that commit is then turned into a patch for the package. If we keep this change in this PR, the patch will contain completely unrelated things.

agraul (changes requested):

salt/modules/aptpkg.py
m-czernek: Ack, I created https://github.com/openSUSE/salt/pull/620 - removing the commit from this PR

agraul (approved):

PR #623, opened 2024-02-05
fix modules/suse_apache missing apache2ctl

M0ses:
The suse_apache module is checking if `apache2ctl` is present at the minion. This check fails on current openSUSE Factory because apache2ctl was renamed to apachectl.

`apache2ctl` isn't execute in the suse_apache module at all.

### What does this PR do?

With this patch the suse_apache module checks for the two really used excecutables (`a2enmod`/`a2dismod`)

Fixes: NONE

### Previous Behavior

enabling/disabling modules failed

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/suse_apache.py
+++ b/salt/modules/suse_apache.py
@@ -19,7 +19,7 @@ def __virtual__():
     """
     Only load the module if apache is installed.
     """
-    if salt.utils.path.which("apache2ctl") and __grains__["os_family"] == "Suse":
+    if salt.utils.path.which("a2enmod") and salt.utils.path.which("a2dismod") and __grains__["os_family"] == "Suse":
         return __virtualname__
     return (False, "apache execution module not loaded: apache not installed.")
 


meaksh:
Hi @M0ses, thanks for contributing. I'm very sorry for the very late response on this one, but we would need an upstream PR for this change (3006.x branch), also including some tests.

Please, see our PR policy: https://github.com/openSUSE/salt/wiki/Development-Environment#github-pull-request-policy

Thanks for understanding!

salt/modules/suse_apache.py
```diff
@@ -19,7 +19,7 @@ def __virtual__():
     """
     Only load the module if apache is installed.
     """
-    if salt.utils.path.which("apache2ctl") and __grains__["os_family"] == "Suse":
+    if salt.utils.path.which("a2enmod") and salt.utils.path.which("a2dismod") and __grains__["os_family"] == "Suse":
```
vzhestkov, line 22: I think maybe better to check this way as we are using the same sources not just for Tumbleweed, but for SLES 15 and Leap 15.x:
```suggestion
    if (
        salt.utils.path.which("apache2ctl") or
        (salt.utils.path.which("a2enmod") and salt.utils.path.which("a2dismod"))
    ) and __grains__["os_family"] == "Suse":
```

vzhestkov (changes requested):

PR #627, opened 2024-02-16
Discover playbooks as either *.yml or *.yaml files (bsc#1211888)

renner:
### What does this PR do?

This patch is modifying the ansiblegate module to detect playbooks by default as either *.yml or *.yaml files. The playbook_extension parameter is changed to receive either a string or a tuple of strings in case you wanted to discover files with different extensions (endswith() can work with both). Even though *.yml seems to be the more popular extension for Yaml files, both extensions seem to be officially valid. Even the [official Ansible documentation](https://docs.ansible.com/ansible/latest/getting_started/get_started_playbook.html#creating-a-playbook) recommends creating a playbook as *.yaml and without this patch it would not be supported out of the box using the ansiblegate module.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1211888

### Previous Behavior

Per default discover_playbooks would detect playbooks only with the *.yml extension. The playbook_extension parameter can receive a different extension as a string, so if you wanted to discover *.yml and *.yaml files you would need to call discover_playbooks twice.

### New Behavior

You can pass a tuple of strings to the playbook_extension parameter, the default value is (".yml", ".yaml").

- [X] Docs
- [X] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/changelog/66048.changed.md
+++ b/changelog/66048.changed.md
@@ -1,0 +1,1 @@
+Ansiblegate discover_playbooks was changed to find playbooks as either *.yml or *.yaml files
--- a/salt/modules/ansiblegate.py
+++ b/salt/modules/ansiblegate.py
@@ -101,17 +101,16 @@ def __virtual__():
 
     proc = subprocess.run(
         [ansible_doc_bin, "--list", "--json", "--type=module"],
-        stdout=subprocess.PIPE,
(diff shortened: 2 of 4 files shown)


m-czernek (commented):
+1, LGTM!

m-czernek (approved):

PR #630, opened 2024-02-21
Decode oscap byte stream to string (bsc#1219001)

m-czernek:
### What does this PR do?

In the openscap module, we send non-escaped byte error stream from the oscap output to salt-master. However, the error stream may contain non-ascii characters, which Uyuni cannot process further. This means that user doesn't see any errors in the UI, just decoding exceptions in the logs.

In this PR, we decode the stream to ascii, and ignore non-ascii characters.

Fixes: Salt-side of https://bugzilla.suse.com/show_bug.cgi?id=1219001

### Commits signed with GPG?
Yes/No


ycedres (approved):
Positive feedback received in the bug report.

PR #636, opened 2024-03-12
Fix tests failures and errors when detected on VM execution from Salt Shaker

meaksh:
### What does this PR do?

This PR fixes a couple of test issues and errors that are raised on our internal Salt Shaker executions on SLE15SP5 when running tests in a VM.

Changes are part of this upstream PR: https://github.com/openSUSE/salt/pull/636

Tracks: https://github.com/SUSE/spacewalk/issues/23286

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/pytests/functional/loader/test_loader.py
+++ b/tests/pytests/functional/loader/test_loader.py
@@ -1,14 +1,23 @@
 import json
+import os
 
 import pytest
 
 from salt.utils.versions import Version
 from tests.support.helpers import SaltVirtualEnv
 from tests.support.pytest.helpers import FakeSaltExtension
+from tests.support.runtests import RUNTIME_VARS
+
+MISSING_SETUP_PY_FILE = not os.path.exists(
+    os.path.join(RUNTIME_VARS.CODE_DIR, "setup.py")
+)
 
 pytestmark = [
     # These are slow because they create a virtualenv and install salt in it
     pytest.mark.slow_test,
+    pytest.mark.skipif(
+        MISSING_SETUP_PY_FILE, reason="This test only work if setup.py is available"
+    ),
 ]
 
 
--- a/tests/pytests/functional/states/test_docker_network.py
+++ b/tests/pytests/functional/states/test_docker_network.py
@@ -220,10 +220,15 @@ def test_present_with_containers(network, docker, docker_network, container):
 
 
 @pytest.mark.parametrize("reconnect", [True, False])
-def test_present_with_reconnect(network, docker, docker_network, container, reconnect):
+def test_present_with_reconnect(
+    network, docker, docker_network, container, reconnect, grains
+):
     """
     Test reconnecting with containers not passed to state
     """
+    if grains["os_family"] == "Suse":
+        pytest.skip("This test is failing for SUSE family")
+
     with network() as net:
         ret = docker_network.present(name=net.name, driver="bridge")
         assert ret.result is True
--- a/tests/integration/pillar/test_git_pillar.py
+++ b/tests/integration/pillar/test_git_pillar.py
@@ -79,6 +79,7 @@ from salt.utils.gitfs import (
     PYGIT2_VERSION,
     FileserverConfigError,
 )
+from salt.utils.versions import Version
 from tests.support.gitfs import (  # pylint: disable=unused-import
     PASSWORD,
     USERNAME,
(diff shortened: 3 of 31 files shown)


tests/pytests/functional/states/test_docker_network.py
```diff
@@ -220,10 +220,15 @@ def test_present_with_containers(network, docker, docker_network, container):
...
 @pytest.mark.parametrize("reconnect", [True, False])
-def test_present_with_reconnect(network, docker, docker_network, container, reconnect):
+def test_present_with_reconnect(
+    network, docker, docker_network, container, reconnect, grains
+):
     """
     Test reconnecting with containers not passed to state
     """
+    if grains["os_family"] == "Suse":
+        pytest.skip("This test is failing for SUSE family")
```
agraul, line 230: Should we skip this for eternity? Why can't we fix the cause of the test failure?

tests/pytests/functional/loader/test_loader.py
```diff
@@ -1,14 +1,23 @@
...
+MISSING_SETUP_PY_FILE = not os.path.exists(
+    os.path.join(RUNTIME_VARS.CODE_DIR, "setup.py")
+)
 
 pytestmark = [
     # These are slow because they create a virtualenv and install salt in it
     pytest.mark.slow_test,
+    pytest.mark.skipif(
+        MISSING_SETUP_PY_FILE, reason="This test only work if setup.py is available"
+    ),
```
agraul, line 20: Can we instead supply a `setup.py`? The loader is a core component, it would be good to test it

agraul (commented):
Looks good overall, just have two questions about skipping specific tests

tests/pytests/functional/loader/test_loader.py
meaksh: It seems it is not enough to provide an example `setup.py` file, neither the `setup.py` from Salt.

These tests are expecting that `salt` is not yet built when running the test (it is expecting the `setup.py` and other build artifacts to be there), as the tests triggers the "salt" building & installation inside a test "virtualenv".

Also the fact that we have splitted our Salt package installation path from our salt testsuite path is also making the tests to mess up.

OTOH these tests are actually running in GH actions, as Salt is not yet built. The tests would be skipped only when running them in a packaged Salt setup.

tests/pytests/functional/states/test_docker_network.py
meaksh: I'm not really sure why exactly this is failing but the error I get is:

```
Failed to create temp network for comparison: Error 500: operation not supported
```

Looking at `docker` service logs in DEBUG:

```
level=debug msg="FIXME: Got an API for which error does not match any expected type!!!" error="operation not supported" error_type=syscall.Errno module=api
level=error msg="Handler for POST /v1.44/networks/create returned error: operation not supported"
level=debug msg="FIXME: Got an API for which error does not match any expected type!!!" error="operation not supported" error_type=syscall.Errno module=api
```

I got the same using SLE15SP4 version of "docker": `24.0.7-ce` and also with latest in Factory: `25.0.3-ce` (but always testing on top of SLE15SP5). Not sure if this is related to the OS, to our docker packages, or really upstream issue.

tests/pytests/functional/loader/test_loader.py
agraul: > OTOH these tests are actually running fine in GH actions, as Salt is not yet built. The tests would be skipped only when running them in a packaged Salt setup.

That by itself is not very convincing, the packaged Salt setup is the setup that's more important to test.

> These tests are expecting that salt is not yet built when running the test (it is expecting the setup.py and other build artifacts to be there), as the tests triggers the "salt" building & installation inside a test "virtualenv".

We might need to revisit this for 3008. From what I can tell, the test covers loading salt extensions and that will be important to cover well.

I'm okay with skipping it for now though. Would be a candidate for the skiplist.

meaksh: Agree, we should ensure these salt extensions tests are running once we go for 3008.

For now, I would rather vote to keep this patch in the codebase instead of adding it to the current `skiplist`, as skiplist will ignore this test in all cases, not only when "setup.py" is missing.

agraul (approved):

PR #643, opened 2024-04-18
Allow explicit reboot on transactional minions

m-czernek:
### What does this PR do?

In this PR, we parse the execution result on a transactional system. If user specified any of the provided `explicit_reboot_cmds` commands in a state file, we issue a reboot even when `activate_transaction` was not explicitly passed to the module.

Note that user still gets the result of `running in chroot, ignoring` (but the system actually reboots).

Part of https://github.com/SUSE/spacewalk/issues/23874

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/modules/transactional_update.py
+++ b/salt/modules/transactional_update.py
@@ -989,8 +989,8 @@ def call(function, *args, **kwargs):
             return local
     finally:
         # Check if reboot is needed
-        if (activate_transaction and pending_transaction()) or _user_specified_reboot(
-            local
+        if (activate_transaction and pending_transaction()) or (
+            not kwargs.get("test", False) and _user_specified_reboot(local)
         ):
             reboot()
 
@@ -1000,13 +1000,13 @@ def _user_specified_reboot(local):
         # Skip if execution is not state/highstate
         return False
 
-    explicit_reboot_cmds = ["reboot", "init 6", "shutdown -r", "shutdown --reboot"]
-    names = []
+    explicit_reboot_cmds = set(["reboot", "system.reboot"])
+    names = set()
     for _, value in local.items():
         if isinstance(value, dict) and "name" in value:
-            names.append(value["name"])
-    # Partial match reboot_cmds to names, so that e.g. "reboot" matches "system.reboot"
-    return any([cmd in name for cmd in explicit_reboot_cmds for name in names])
+            names.add(value["name"])
+
+    return bool(explicit_reboot_cmds.intersection(names))
 
 
 def apply_(mods=None, **kwargs):
--- a/salt/modules/transactional_update.py
+++ b/salt/modules/transactional_update.py
@@ -283,6 +283,7 @@ import sys
 import salt.client.ssh.state
 import salt.client.ssh.wrapper.state
 import salt.exceptions
+import salt.executors.transactional_update
 import salt.utils.args
 from salt.modules.state import _check_queue, _prior_running_states, _wait, running
 
@@ -996,14 +997,15 @@ def call(function, *args, **kwargs):
 
 
 def _user_specified_reboot(local, function):
-    if function != "state.highstate" and function != "state.sls":
+    explicit_reboot_cmds = set(["reboot", "system.reboot"])
+    explicit_reboot_modules = ["cmd", "module"]
+
+    if function not in salt.executors.transactional_update.DELEGATION_MAP.keys():
         return False
 
     if not isinstance(local, dict):
         return False
 
-    explicit_reboot_cmds = set(["reboot", "system.reboot"])
-    explicit_reboot_modules = ["cmd", "module"]
     names = set()
     for execution_id, execution_result in local.items():
         if not isinstance(execution_id, str):
--- a/salt/modules/transactional_update.py
+++ b/salt/modules/transactional_update.py
@@ -985,11 +985,28 @@ def call(function, *args, **kwargs):
             else:
                 return local
         except ValueError:
-            return {"result": False, "retcode": 1, "comment": ret_stdout}
+            local = {"result": False, "retcode": 1, "comment": ret_stdout}
+            return local
     finally:
         # Check if reboot is needed
-        if activate_transaction and pending_transaction():
+        if (activate_transaction and pending_transaction()) or _user_specified_reboot(
+            local
+        ):
             reboot()
+
+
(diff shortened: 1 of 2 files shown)


salt/modules/transactional_update.py
```diff
@@ -985,13 +985,30 @@ def call(function, *args, **kwargs):
...
 
+def _user_specified_reboot(local):
+    if not isinstance(local, dict):
+        # Skip if execution is not state/highstate
+        return False
+
+    explicit_reboot_cmds = ["reboot", "init 6", "shutdown -r", "shutdown --reboot"]
+    names = []
+    for _, value in local.items():
+        if isinstance(value, dict) and "name" in value:
```
agraul, line 1006: Does this match all kinds of state functions? If so, we risk rebooting when it wasn't actually requested. I think we should prefer false-negatives over false-positives. Rebooting when it's not asked for is disruptive while a missing reboot can be worked around.

salt/modules/transactional_update.py
```diff
@@ -985,13 +985,30 @@ def call(function, *args, **kwargs):
...
+            names.append(value["name"])
+    # Partial match reboot_cmds to names, so that e.g. "reboot" matches "system.reboot"
+    return any([cmd in name for cmd in explicit_reboot_cmds for name in names])
```
agraul, line 1009: We also need to check if this was called by `state.test` or with `test=True`

salt/modules/transactional_update.py
```diff
@@ -985,13 +985,30 @@ def call(function, *args, **kwargs):
...
+        if (activate_transaction and pending_transaction()) or _user_specified_reboot(
+            local
+        ):
             reboot()
 
 
+def _user_specified_reboot(local):
+    if not isinstance(local, dict):
+        # Skip if execution is not state/highstate
+        return False
```
agraul, line 1001: This does skip if this is not a state application, but it will also match arbitrary execution modules that return dicts.

agraul (changes requested):
It's good to have this as a first try. As you said, it already works. The risk with the current state is that it works too often, we should be more defensive :smile:

salt/modules/transactional_update.py
m-czernek: I've modified the actual matching below since this comment; so while we do iterate over all state functions, the function is matched with exact match only. Right now, we only match names that equal to `reboot` or `system.reboot` (from `module.include`). Is that sufficient to alleviate your concern, or do you think we should also match the key? (e.g. `cmd` or `module` in `key`)

agraul: > Side note: the following sls seems not to start a transaction, so this change doesn't affect it.

Is that really the case? Without this patch, the condition to decide if a reboot should be called requires a pending transaction. With this patch, it's either pending transaction + activate_transaction=True or we find `name: reboot`. Am I missing something?

> The same applies to state.test - it also doesn't run in a transaction.
Right, that's the default. We might want to still rule that out (as you've done anyway), the configuration what runs inside/outside a transaction can be configured.

m-czernek: No indeed, you were right. When I was testing it, I think I mistakenly turned off the t-u executor. Upon re-checking my results, I realized I was wrong, and added a further check

salt/modules/transactional_update.py
```diff
@@ -985,13 +985,40 @@ def call(function, *args, **kwargs):
...
 
+def _user_specified_reboot(local, function):
+    if function != "state.highstate" and function != "state.sls":
+        return False
+
+    if not isinstance(local, dict):
+        return False
+
+    explicit_reboot_cmds = set(["reboot", "system.reboot"])
+    explicit_reboot_modules = ["cmd_", "module_"]
```
vzhestkov, line 1006: Maybe it's a bit wrong to specify here `_` as a part of the module name, actually `_|-` is used as a separator in the state apply results like you use in the tests:
```
            "cmd_|-reboot_test_|-reboot_|-run": {
                "name": "reboot",
                "changes": {},
                "result": True,
            }
```
I don't know why but actually, `cmd_|-reboot_test_|-reboot_|-run` is a `module name`, `id`, `name`, `fun` separated with `_|-`, so maybe it's better to parse it using `_|-` instead of relying on `|`, but there is a mess that actually either the `name` or `id` can also contain `_|-` and it could add the mess, but even existing code in salt is producing wrong output in such cases.

salt/modules/transactional_update.py
```diff
@@ -985,13 +985,40 @@ def call(function, *args, **kwargs):
...
+    names = set()
+    for full_module_name, module_result in local.items():
+        if not isinstance(full_module_name, str):
+            continue
+
+        module = full_module_name.split("|")[0]
```
vzhestkov, line 1012: `|` here is the separator I mentioned in the previous comment.

vzhestkov (changes requested):
I think it's better to adjust the code for `_|-` as a separator.

salt/modules/transactional_update.py
m-czernek: Makes sense,thank you for the suggestion! Implemented.

salt/modules/transactional_update.py
```diff
@@ -985,13 +985,40 @@ def call(function, *args, **kwargs):
...
         # Check if reboot is needed
-        if activate_transaction and pending_transaction():
+        if (activate_transaction and pending_transaction()) or (
+            not kwargs.get("test", False) and _user_specified_reboot(local, function)
+        ):
             reboot()
 
 
+def _user_specified_reboot(local, function):
+    if function != "state.highstate" and function != "state.sls":
```
vzhestkov, line 999: Looks like we need to cover all of these functions: https://github.com/openSUSE/salt/blob/86d6235df59936c038df608030a715901897b85b/salt/executors/transactional_update.py#L14-L19
Actually `state.highstate` is not widely used in Uyuni/SUMA, but mostly `state.apply` with no `mods` specified (as `state.apply` with no mods is acting as `state.highstate`)

m-czernek: Indeed, great catch - didn't want to maintain the list of functions in two places, so I pulled in the `DELEGATION_MAP` directly. WTYD?

vzhestkov (approved):
LGTM, let's try

PR #651, opened 2024-05-13
Change zfs module test dependency from man to more

m-czernek:
### What does this PR do?
The test requires a file that the test can open and read. The problem with temp files is that the files pytest keeps the files open during the test (because the file gets deleted when closed). Consequently, the test previously used the `/usr/bin/man` file.

With our support of SL Micro, `man` is not available, so I changed `man` to `more`.


m-czernek:
Test failure resolved without the change, closing

PR #656, opened 2024-06-06
Provide systemd timer unit

m-czernek:
### What does this PR do?

Provide systemd timer unit that activates the cleanup script once, 60s after the venv-salt-minion package is installed. Spec changes will come in OBS, where we maintain the venv-salt-minion spec file.


meaksh (approved):

meaksh:
It seems the current test failure is caused by a new version of `requests`: https://github.com/docker/docker-py/issues/3256

PR #659, opened 2024-06-17
Fix "status.diskusage" and exclude some tests to run when testing Salt Bundle

meaksh:
### What does this PR do?
This PR fixes an issue with `status.diskusage` function and also exclude tests that are failing when running with Salt Bundle, as tests needs to be adapted.

Since the same tests are working fine for the classic package, we do not exclude them completely via `skiplist`.

- Upstream PR: https://github.com/saltstack/salt/pull/66647 (Only some commits from this PR make sense upstream)

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/status.py
+++ b/salt/modules/status.py
@@ -1061,6 +1061,7 @@ def diskusage(*args):
                 ret[path] = {"available": available, "total": total}
             except OSError as exc:
                 log.warning("Cannot get stats from '{}': {}".format(path, exc))
+                ret[path] = {"available": None, "total": None}
     return ret
 
 
--- a/salt/modules/status.py
+++ b/salt/modules/status.py
@@ -1053,11 +1053,14 @@ def diskusage(*args):
     ret = {}
     for path in selected:
         if os.path.exists(path):
-            fsstats = os.statvfs(path)
-            blksz = fsstats.f_bsize
-            available = fsstats.f_bavail * blksz
-            total = fsstats.f_blocks * blksz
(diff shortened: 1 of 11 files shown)


salt/modules/status.py
```diff
@@ -1053,11 +1053,14 @@ def diskusage(*args):
...
-            total = fsstats.f_blocks * blksz
-            ret[path] = {"available": available, "total": total}
+            try:
+                fsstats = os.statvfs(path)
+                blksz = fsstats.f_bsize
+                available = fsstats.f_bavail * blksz
+                total = fsstats.f_blocks * blksz
+                ret[path] = {"available": available, "total": total}
+            except OSError as exc:
+                log.warning("Cannot get stats from '{}': {}".format(path, exc))
```
m-czernek, line 1063: So personally, I would expect something like:

```suggestion
                log.warning("Cannot get stats from '{}': {}".format(path, exc))
                ret[path] = {"available": None, "total": None}
```
Or something similar to let the user know there was an error right away with the path. But, not sure, if upstream is OK with it, why not.

m-czernek (approved):
Leaving a minor non-blocking comment - otherwise LGTM.

salt/modules/status.py
meaksh: Thanks for the comment. Indeed, it sounds reasonable, and I also like it more than the current approach.

I'll include your suggestion here and also to the upstream PR :+1:

PR #672, opened 2024-08-21
Improve error handling with different OpenSSL versions

vzhestkov:
### What does this PR do?

Backports https://github.com/saltstack/salt/pull/66818

With the most recent versions of `cryptography` module the exception value which is checked here https://github.com/saltstack/salt/blob/246d0664577ef72da8bd1f0c4dff0d18b4428b23/salt/utils/x509.py#L704 is different.
The latest version of `cryptography` is returning https://github.com/pyca/cryptography/blob/932b8a3f67810140a6e178f7b676e1cb9c3585b1/src/rust/src/backend/utils.rs#L463

It could also be returned with the lower version of `cryptography` depending on the combination with the OpenSSL version it's used with.

Tracks: https://github.com/SUSE/spacewalk/issues/24859

### Previous Behavior
`x509.private_key_managed` state function could fail with the comment `Could not load PEM-encoded private key`
The following tests could fail as well:
```
tests/pytests/functional/states/test_x509_v2.py::test_private_key_managed_passphrase_changed_overwrite
tests/pytests/functional/states/test_x509_v2.py::test_private_key_managed_passphrase_changed_not_overwrite
```

### New Behavior
No test fails and `x509.private_key_managed` state with most recent `cryptography` or some other OpenSSL versions which can produce different errors on such cases.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/utils/x509.py
+++ b/salt/utils/x509.py
@@ -695,7 +695,8 @@ def load_privkey(pk, passphrase=None, get_encoding=False):
                 return pk, "pem", None
             return pk
         except ValueError as err:
-            if "Bad decrypt" in str(err):
+            str_err = str(err)
+            if "Bad decrypt" in str_err or "Could not deserialize key data" in str_err:
                 raise SaltInvocationError(
                     "Bad decrypt - is the password correct?"
                 ) from err
--- a/tests/pytests/functional/states/test_x509_v2.py
+++ b/tests/pytests/functional/states/test_x509_v2.py
@@ -3,6 +3,8 @@ from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


vzhestkov:
Commit https://github.com/openSUSE/salt/pull/672/commits/f0a69e927103ae793a41fc6da90fedc08466b10b is different from the upstream PR, but it only makes sense for outdated versions of OpenSSL, like the ones used in CentOS7 and SLE12, so it doesn't make any sense to try to push it upstream.

m-czernek (approved):
+1, LGTM

PR #684, opened 2024-10-03
Fix x509 private key tests and test_suse on SLE12

m-czernek:
### What does this PR do?

This PR is related to https://github.com/openSUSE/salt/pull/682.

For x509 tests, I forgot to include two tests.

For the `test_suse.py` test, the problem there is that on SLE12, none of the if/else paths matched. I simplified it so that we use the SLE15 SP4 repo URL, since the tests check whether we can manipulate with repositories (e.g. add, modify, remove, ...) but does not install any packages (so the repository URL shouldn't matter, we just need to provide one)

Relates to: https://github.com/SUSE/spacewalk/issues/23286

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


vzhestkov (approved):
LGTM

PR #688, opened 2024-10-31
Enhance find_json garbage filtering (bsc#1231605)

m-czernek:
### What does this PR do?

In this PR, we enhance the `find_json` function to work with valid JSONs that contain garbage right after the JSON end. This is a problem when the text contains JSON and, for example, deprecation messages, such as:

```
{
    "local": {
        "saltutil_|-sync_states_|-sync_states_|-sync_states": {
            "name": "sync_states",
            "changes": {},
            "result": true,
            "comment": "No updates to sync",
            "__sls__": "util.syncstates",
            "__run_num__": 0,
            "start_time": "14:35:37.150819",
            "duration": 1256.999,
            "__id__": "sync_states"
        },
...JSON continues...
}/usr/lib/python3.6/site-packages/salt/states/x509.py:214: DeprecationWarning: The x509 modules are deprecated. Please migrate to the replacement modules (x509_v2). They are the default from Salt 3008 (Argon) onwards.
  "The x509 modules are deprecated. Please migrate to the replacement "
```

Related to: [#25514](https://github.com/SUSE/spacewalk/issues/25514)
Upstream PR: https://github.com/saltstack/salt/pull/67023

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -39,6 +39,7 @@ def find_json(raw):
     # Search for possible starts end ends of the json fragments
     for ind, _ in enumerate(lines):
         line = lines[ind].lstrip()
+        line = line[0] if line else line
         if line == "{" or line == "[":
             starts.append((ind, line))
         if line == "}" or line == "]":
@@ -61,10 +62,17 @@ def find_json(raw):
         working = "\n".join(lines[start : end + 1])
         try:
             ret = json.loads(working)
+            return ret
+        except ValueError:
+            pass
+        # Try filtering non-JSON text right after the last closing curly brace
+        end_str = lines[end].lstrip()[0]
+        working = "\n".join(lines[start : end]) + end_str
+        try:
+            ret = json.loads(working)
+            return ret
         except ValueError:
             continue
-        if ret:
-            return ret
 
     # Fall back to old implementation for backward compatibility
     # excpecting json after the text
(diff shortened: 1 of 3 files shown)


salt/utils/json.py
```diff
@@ -39,6 +39,7 @@ def find_json(raw):
     # Search for possible starts end ends of the json fragments
     for ind, _ in enumerate(lines):
         line = lines[ind].lstrip()
+        line = line[0] if line else line
```
vzhestkov, line 42: I'm bit confused with this part, maybe there are no such examples, but for me it looks like it can work with only pretty formated json. It's confusing because looks like it just takes only first symbol from the string with ignoring the rest part completely.

m-czernek: It does work with pretty formatted JSONs only indeed. I don't think `json.find_json` supports unformatted JSONS - except for the best-effort search at the end, which just filters the text before it and then tries to load whatever is left.

Our implementation that you authored definitely doesn't support unformatted JSONs:

```
    for ind, _ in enumerate(lines):
        line = lines[ind].lstrip()
        line = line[0] if line else line
        if line == "{" or line == "[": # <-- this will never work for unformatted JSON
            starts.append((ind, line))
        if line == "}" or line == "]": # <-- this will never work for unformatted JSON
            ends.append((ind, line))
```

In other words, if I'm not mistaken, we didn't support unformatted JSONs before, and we still don't, right?

meaksh (approved):

vzhestkov (approved):
LGTM

PR #696, opened 2025-01-20
Make `_auth` calls visible with master stats

vzhestkov:
### What does this PR do?

`_auth` calls are insivisble in case of using `auth_events: False` in master's config and even with using `master_stats: True` these calls are not listed there as performed before the code reachin the part which is calculating statistics.

Upstream PR: https://github.com/saltstack/salt/pull/67746
Tracks: https://github.com/SUSE/spacewalk/issues/24400

### Previous Behavior
`_auth` calls are invisible in master stats

### New Behavior
`_auth` calls are listed and populated with master stats enabled

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -1036,7 +1036,7 @@ class MWorker(salt.utils.process.SignalHandlingProcess):
 
         :param dict payload: The payload route to the appropriate handler
         """
-        if payload["cmd"] == "_auth":
+        if payload.get("cmd") == "_auth":
             if self.opts["master_stats"]:
                 self.stats["_auth"]["runs"] += 1
                 self._post_stats(payload["_start"], "_auth")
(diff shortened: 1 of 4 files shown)


salt/master.py
```diff
@@ -1036,6 +1036,11 @@ def _handle_payload(self, payload):
 
         :param dict payload: The payload route to the appropriate handler
         """
+        if payload.get("cmd") == "_auth":
+            if self.opts["master_stats"]:
```
meaksh, line 1040: Should we use:
```suggestion
            if self.opts.get("master_stats", False):
```

to be on the safe side, as we do in `salt/channel/server.py`?

vzhestkov: with `master.py` there are another places which are relying on the presence of these parameter in the `opts`, the other master stats related conditions are using `self.opts["master_stats"]`, originally I was using exactly `self.opts.get("master_stats", False)`, but on preparing upstream PR changed it to the original way of checking this parameter.

meaksh: ACK. Thanks for claryfing

meaksh (approved):

PR #706, opened 2025-02-26
Implement multiple inventory for `ansible.targets`

vzhestkov:
### What does this PR do?

Adds `inventories` parameter for `ansible.targets` in case if the list passed to this parameter the function will return the dict with the inventory filenames as the keys and the content inside.

Tracks: https://github.com/SUSE/spacewalk/issues/26439
Upstream PR: https://github.com/saltstack/salt/pull/67776

### Previous Behavior
`ansible.targets` returns the content of only one single inventory

### New Behavior
`ansible.targets` can return the content of multiple inventory files

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/utils/ansible.py
+++ b/salt/utils/ansible.py
@@ -18,32 +18,55 @@ def __virtual__():  # pylint: disable=expected-2-blank-lines-found-0
     return (False, "Install `ansible` to use inventory")
 
 
-def targets(inventory="/etc/ansible/hosts", yaml=False, export=False):
+def targets(inventory=None, inventories=None, yaml=False, export=False):
     """
     Return the targets from the ansible inventory_file
     Default: /etc/salt/roster
     """
-    if not os.path.isfile(inventory):
-        raise CommandExecutionError("Inventory file not found: {}".format(inventory))
-    if not os.path.isabs(inventory):
-        raise CommandExecutionError("Path to inventory file must be an absolute path")
+
+    if inventory is None and inventories is None:
+        inventory = "/etc/ansible/hosts"
+    multi_inventory = True
+    if not isinstance(inventories, list):
+        multi_inventory = False
+        inventories = []
+    if inventory is not None and inventory not in inventories:
+        inventories.append(inventory)
 
     extra_cmd = []
     if export:
         extra_cmd.append("--export")
     if yaml:
         extra_cmd.append("--yaml")
-    inv = salt.modules.cmdmod.run(
-        "ansible-inventory -i {} --list {}".format(inventory, " ".join(extra_cmd)),
-        env={"ANSIBLE_DEPRECATION_WARNINGS": "0"},
-        reset_system_locale=False,
-    )
-    if yaml:
-        return salt.utils.stringutils.to_str(inv)
-    else:
-        try:
-            return salt.utils.json.loads(salt.utils.stringutils.to_str(inv))
-        except ValueError:
-            raise CommandExecutionError(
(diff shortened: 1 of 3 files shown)


salt/utils/ansible.py
```diff
@@ -18,32 +18,55 @@ def __virtual__():  # pylint: disable=expected-2-blank-lines-found-0
...
-        raise CommandExecutionError("Path to inventory file must be an absolute path")
+
+    if inventory is None and inventories is None:
+        inventory = "/etc/ansible/hosts"
+    multi_inventory = True
+    if not isinstance(inventories, list):
+        multi_inventory = False
+        inventories = []
+    if inventory is not None and inventory not in inventories:
+        inventories.append(inventory)
```
meaksh, line 34: Hmm, even if functional, it feels a bit strange that we allow passing both `inventory` and `inventories` parameters at the same.

Also, we are breaking the expected format for the output of the already existing `ansible.targets` function when passing a single `inventory`. I would want to prevent this as possible, so any automation already done by a customer relying on the previous output format is not breaking now.

So, my approach would be:
- if calling `ansible.targets(inventory="/foo/bar")` -> produces the same output as before, only containing a single inventory and not a list of inventories as output.
- if calling `ansible.targets(inventories=["/foo/bar", "/foo/bar2"])` -> produces a new output format, containing a list of inventories.
- if calling `ansible.targets(inventory="/foo/deadbeaf", inventories=["/foo/bar", "/foo/bar2"])`  -> raises exception, as you cannot use both `inventory` and `inventories` at the same time.

meaksh (changes requested):

salt/utils/ansible.py
vzhestkov: Maybe it's not that clear from the code, but in case of passing just single `inventory` the behaviour is not changing at all, but it's changing in case if `inventories` are specified as the list and in case of specifying both if the `inventory` is not listed in the `inventories` it will be included to the list. So the behaviour here is changing only in case of presence of `inventories`. From your suggested approach there is only one difference now, the latest point, we are not raising the exception but just extending the inventories list with the value from inventory.

meaksh: Right, sorry for the confusion, I missed this: :sweat_smile: 

```
if not multi_inventory:
    ret = inv
    break
```

The last point still makes sense to me though, I wouldn't allow the usage of the two arguments at the same time. What do you think?

vzhestkov: I don't think it should be a common case, but why not to give some extra flexibility, I thought if someone will specify the both parameters they can do it on purpose or maybe by accident, but if the specified file is already listed in `inventories` it shouldn't be a problem as well as the case when someone will keep one of the inventories separate from the others, I don't know why it could be, but I don't want to make it too strict in the places where it could be redundant.

meaksh: I don't really have a very strong opinion here. I agree allowing both gives more flexibility but it is a strange and messy case, because if the users want to pass more than one inventory, then they should use `inventories` and not mixing those two arguments. It feels to me like we are tolerating here the case when the user is messing with their arguments.

I don't really see any valid case for a user to use both parameters at the same time, do  you see any? to me, it feels more natural that both functionalities/outputs are kept separately.

vzhestkov: I would prefer not to be that strict where it's not really required and there are 2 approves from upstream for such implementation already, so I would keep it as it is.

meaksh (approved):

PR #711, opened 2025-03-31
Use internal salt.utils.pkg.deb classes instead of aptsource

vzhestkov:
### What does this PR do?

Upstream PR: https://github.com/saltstack/salt/pull/67956

Instead of using `aptsources` from Debian distros, use internal `salt.utils.pkg.deb` classes to avoid extra dependency.

Tracks: https://github.com/SUSE/spacewalk/issues/25989

### Previous Behavior
Possible fails on some new Debian based distros like Ubuntu 24.04 due to missing implementation for deb822 with the internals of `salt.utils.pkg.deb`

### New Behavior
No fails on managing apt sources of deb822 format

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No


salt/modules/aptpkg.py
```diff
@@ -18,7 +16,6 @@
 import pathlib
 import re
 import shutil
-import tempfile
 import time
 from urllib.error import HTTPError
 from urllib.request import Request as _Request
```
m-czernek, line 21: L24 and 25 now have unused imports that we should delete.

m-czernek (approved):
This looks good to me. In my testing, I looked at both Deb822 repos, as well as the old one-line style repos which e.g. stopped working previously in the QE tests. Though I could have a few comments about the deb822-related code, I don't think it makes a lot of sense since we took it from python3-apt.

The main note here is about the license, as we discussed previously. We should ensure that we leave a note in the `deb.py` class that this code is from `python3-apt`. Then, we might want to wait with the merge until we know what upstream does with the code.

I leave that up to you - if we merge now, we might have to revert this later in case upstream rejects this. If upstream rejects this, we might want to create this as a separate package due to license, and again revert this. It would make sense to me to do this upstream-first. 

But functionally, this PR looks good to me. Thank you for looking into this!

salt/modules/aptpkg.py
vzhestkov: Not sure if I get the point regarding unused imports, which ones you mean? `_Request` and `_urlopen`. If so, they are still used.

meaksh (changes requested):
I see some tests failures for this PR. Let's fix them before merging.

salt/modules/aptpkg.py
```diff
@@ -2802,40 +2485,29 @@ def mod_repo(repo, saltenv="base", aptkey=True, **kwargs):
...
+    full_comp_list = [comp.strip() for comp in repo_entry["comps"]]
     no_proxy = __salt__["config.option"]("no_proxy")
 
-    kwargs["signedby"] = pathlib.Path(repo_signedby) if repo_signedby else ""
-
-    if not aptkey and not kwargs["signedby"]:
-        raise SaltInvocationError("missing 'signedby' option when apt-key is missing")
+    kwargs["signedby"] = (
+        pathlib.Path(repo_entry["signedby"]) if repo_entry["signedby"] else ""
+    )
```
m-czernek, line 2508: [non-blocking comment]

Note, these lines mean that in a SLS like this:

```sls
manage_noble_security_repo:
  pkgrepo.managed:
    - name: "deb http://security.ubuntu.com/ubuntu noble-security main universe restricted multiverse"
    - file: /etc/apt/sources.list.d/myrepo.sources
    - signedby: /usr/share/keyrings/ubuntu-archive-keyring.gpg
```

Salt ignores the `signedby` kwarg. If we added these lines before the highlighted lines, it would still propagate to the repo file:

```python
if 'signedby' in kwargs and not repo_entry["signedby"]:
    repo_entry["signedby"] = kwargs["signedby"]
```

It is a bit confusing to me because the code uses a number of `kwargs["signedby"]` but we null it in case it does not appear in the name, like `deb [signed-by=/etc/apt/keyrings/salt-archive-keyring.gpg] ...`

This means that we cannot modify the `signedby` data with the kwargs, like we do for e.g. suites, uris, etc - we have to modify the `name` argument instead..

salt/modules/aptpkg.py
m-czernek: I mean:

```python
import salt.config
import salt.syspaths
```

They seem not to be used by the code, can we delete those?

tests/pytests/functional/modules/test_pkg.py
```diff
@@ -219,12 +219,15 @@ def test_owner(modules, grains):
...
 @pytest.mark.skip_on_freebsd(reason="test for new package manager for FreeBSD")
 @pytest.mark.requires_salt_modules("pkg.which")
-def test_which(modules):
+def test_which(grains, modules):
     """
     test finding the package owning a file
     """
-    func = "pkg.which"
-    ret = modules.pkg.which("/usr/local/bin/salt-call")
+    if grains["os_family"] in ["Debian", "RedHat"]:
```
m-czernek, line 226: Just wondering here, why don't we take the code from https://github.com/saltstack/salt/blob/master/tests/pytests/functional/modules/test_pkg.py#L233 ?

salt/utils/pkg/deb.py
```diff
@@ -2,6 +2,804 @@
...
+            + [cdrom_match.group(2).strip()]
+            + [p.strip() for p in cdrom_match.group(3).split()]
+        )
+    else:
+        repo_line = line.strip().split()
+    if (
+        not repo_line
+        or repo_line[0] not in ["deb", "deb-src", "rpm", "rpm-src"]
+        or len(repo_line) < 3
+    ):
```
m-czernek, line 766: I don't think `len(repo_line) < 3` is a sufficient condition here. Consider the following valid name:

```
"deb [trusted=yes] http://minima-example.de/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/deb/ /" 
``` 

This works, but because of the `trusted=yes` part, `len(repo_line)` is `4`.

Now if we consider the following invalid name:

```
"deb [trusted=yes] http://minima-example.de/repositories/systemsmanagement:/Uyuni:/Test-Packages:/Pool/deb/"
```
(it is missing the suite, the last `/`), the  `len(repo_line)` is `3`, so it passes the check. However, we get the following stack trace:

```
...
                File "/usr/lib/venv-salt-minion/lib/python3.11/site-packages/salt/modules/aptpkg.py", line 1728, in get_repo
                  repo_entry = _split_repo_str(repo)
                               ^^^^^^^^^^^^^^^^^^^^^
                File "/usr/lib/venv-salt-minion/lib/python3.11/site-packages/salt/modules/aptpkg.py", line 1493, in _split_repo_str
                  entry = SourceEntry(repo)
                          ^^^^^^^^^^^^^^^^^
                File "/usr/lib/venv-salt-minion/lib/python3.11/site-packages/salt/utils/pkg/deb.py", line 319, in __init__
                  self.parse(line)
                File "/usr/lib/venv-salt-minion/lib/python3.11/site-packages/salt/utils/pkg/deb.py", line 398, in parse
                  self.dist = repo_line[2]
                              ~~~~~~~~~^^^
              IndexError: list index out of range

```

salt/modules/aptpkg.py
```diff
@@ -1722,56 +1480,22 @@ def version_cmp(pkg1, pkg2, ignore_epoch=False, **kwargs):
...
-    return (
-        split.type,
-        split.architectures,
-        split.uri,
-        split.dist,
-        split.comps,
-        signedby,
-    )
+    entry = SourceEntry(repo)
+    invalid = entry.invalid
```
m-czernek: This line is unused, did you want to use it or shall we remove it?

salt/modules/aptpkg.py
```diff
@@ -1722,56 +1480,22 @@ def version_cmp(pkg1, pkg2, ignore_epoch=False, **kwargs):
...
+    signedby = entry.signedby
+
+    return {
+        "type": entry.type,
+        "architectures": entry.architectures,
+        "uri": entry.uri,
+        "dist": entry.dist,
+        "comps": entry.comps,
+        "signedby": signedby,
```
m-czernek, line 1494: We're actually missing the `trusted` parameter here. If `trusted=yes` is set in the opts, like `deb [trusted=yes] ...`, this results in no change to the generated repo file.

It should result in:

```
Types: deb
URIs: ...
Suites: ...
...
Trusted: yes
```

salt/modules/aptpkg.py
m-czernek: Removed

salt/modules/aptpkg.py
m-czernek: Fixed

tests/pytests/functional/modules/test_pkg.py
m-czernek: Leaving for later review.

salt/modules/aptpkg.py
m-czernek: Removed

salt/modules/aptpkg.py
m-czernek: Leaving for later review, since we don't document `signedby` kwarg anywhere. This is a potential improvement for a future PR.

salt/utils/pkg/deb.py
m-czernek: Improved error when the user inputs incorrect deb line.

m-czernek (approved):

PR #714, opened 2025-04-02
Pin GitHub actions to commits

ycedres:
### What does this PR do?
Following [Security hardening for GitHub Actions guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions?learn=getting_started) actions should be pinned to a full length commit SHA. This PR changes action tags to their corresponding commit.

Fixes: https://github.com/SUSE/spacewalk/issues/26740

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


meaksh (changes requested):
Thanks for the changes. There are few things we should take care:
- The commit message must include a `[skip]` prefix, so we avoid creating a new patch file for these changes that are not done to the salt codebase.
- The files under `.github/actions/` are not really used, as no workflow are using them here. We can just remove all those files instead of fixing them.

ycedres:
> Thanks for the changes. There are few things we should take care:
> 
> * The commit message must include a `[skip]` prefix, so we avoid creating a new patch file for these changes that are not done to the salt codebase.
> * The files under `.github/actions/` are not really used, as no workflow are using them here. We can just remove all those files instead of fixing them.

ln order to remove files under `.github/actions` I'd like to make sure we won't need in the future the usage of some of the actions that apparently is done here:https://github.com/openSUSE/salt/blob/722484c93995b253bc61e93867a521693d3ca0b3/tools/ci.py#L277

Probably just used by SaltStack CI but it will be not consistent in any case after file removal.

meaksh:
> > Thanks for the changes. There are few things we should take care:
> > 
> > * The commit message must include a `[skip]` prefix, so we avoid creating a new patch file for these changes that are not done to the salt codebase.
> > * The files under `.github/actions/` are not really used, as no workflow are using them here. We can just remove all those files instead of fixing them.
> 
> ln order to remove files under `.github/actions` I'd like to make sure we won't need in the future the usage of some of the actions that apparently is done here:
> 
> https://github.com/openSUSE/salt/blob/722484c93995b253bc61e93867a521693d3ca0b3/tools/ci.py#L277
> 
> Probably just used by SaltStack CI but it will be not consistent in any case after file removal.

We don't really use any of those actions anymore in our openSUSE repo.

meaksh (approved):

PR #730, opened 2025-10-01
Add python3.11 as preferable for salt-ssh to avoid tests fails

vzhestkov:
### What does this PR do?

With `salt-ssh` the shim is used on the client side. With `Uyuni` or `SUSE Multi Linux Manager` we are using Salt Bundle only for now, so it doesn't make any sense to try to run the shim with the interpreter which is in EoL for a long time already.
For testing it should be fine to choose python 3.11 as preferred one.

Tracks: https://github.com/SUSE/spacewalk/issues/26171

### Previous Behavior
Some salt-ssh related tests were failing as salt ssh shim was running in the testing environment with python 3.6 even if the flavor of test was selected as python 3.11

### New Behavior
No failing tests due to the issue of selecting wrong interpreter

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -157,7 +157,7 @@ SSH_PY_CODE='import base64;
 if [ -n "$DEBUG" ]
     then set -x
 fi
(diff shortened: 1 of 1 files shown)


meaksh (approved):
LGTM, thanks!

PR #731, opened 2025-10-03
Allow libgit2 to guess sysdir homedir successfully (bsc#1250520) (bsc#1227207)

meaksh:
### What does this PR do?

This PR fixes the generic error that is still seen around gitfs with certain `pygit2` and `libgit2` versions:

```
_pygit2.GitError: error loading known_hosts:
```

After debugging `libgit2` and `pygit2` I was able to understand what's going on here. The internal `git_sysdir__dirs` struct is populated during `libgit2` initialization, which is only triggered at the time of importing `pygit2` module.

If the `HOME` environment variable is not defined when importing `pygit2`, then the internal guess function from `libgit2` to determine the homedir (`git_sysdir_guess_home_dirs`) fails to define the HOME path for the initializated stack. Therefore the above error occurs.

Also once `pygit2/libgit2` is instantiated, there is no way until `pygit2` version 1.18.2 to explicitely redefine the internal home dir on the initializated stack, as `GIT_OPT_SET_HOMEDIR` wasn't implemented in `pygit2` until 1.18.2. See https://github.com/libgit2/pygit2/pull/1409

So, this PR makes sure the `HOME` env variable is present at the time of importing `pygit2` at `salt.utils.gitfs`.

Upstream PR: https://github.com/saltstack/salt/pull/68366

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -110,6 +110,15 @@ try:
 
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
+        if "HOME" not in os.environ:
+            # Make sure $HOME env variable is set before importing pygit2 to prevent
+            # _pygit2.GitError: error loading known_hosts in some libgit2 versions.
+            # The internal "git_sysdir__dirs" from libgit2, is initializated
+            # when importing pygit2. The $HOME env must be present to allow libgit2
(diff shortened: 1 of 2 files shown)


vzhestkov (approved):
LGTM :+1:

PR #736, opened 2025-10-29
Fix Salt for Python >3.11

m-czernek:
### What does this PR do?

In this PR, we migrate Salt to external tornado when the system uses Python greater than 3.11 (i.e. 3.12 and newer). This is specifically a problem on Tumbleweed, where Salt does not work with the old vendored Tornado. 

Cherry-picked commits:

- 551c39b09732470474919dc143897aec1f94df43
- 8db1291eced8de112bbb1c9655a20ce9fcce9971
- 308958fa90bce1541a7fd364eaa2cb9409e780d6
- 2b38571f3998068a47e780183e812bbb577cb93b
- 03c8964e606389436b7f93d4b55dff9e9a850179
- 7264a2810d83bedb35091e862ba30934801e3b80
- e7b625596b7c3ca8138aeeeeda2c995276fba964
- 49d1dd5d9f552554ffb9dbb266715dc45a25ba19
- 00ae0862a76e3ae32f8901527811bf538b30020f
- c4677c6eb4532212406cdffd935e66a97c68490f (partial)
- 44314d663604981b087fd7c335ef16c9ad475d42
- 0786d7a8433787d1ebce65a5a62d1d6213c489f5

Additional commits provide fixes so that Salt can select which tornado to use based on used Python version, as well as some minor fixes to tests (so that the github testsuite stays green).

Related to: https://github.com/SUSE/spacewalk/issues/26953

### Previous Behavior
Salt is broken on OSes that use Python 3.13.

### New Behavior
Salt works on OSes that use Python 3.13

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


salt/utils/thin.py
```diff
@@ -24,7 +24,7 @@
 
 import salt
 import salt.exceptions
-import salt.ext.tornado as tornado
+import tornado as tornado
```
m-czernek: TODO (waiting for review to be done before making another commit)

```suggestion
import tornado
```

salt/client/mixins.py
```diff
@@ -379,29 +378,27 @@ def low(self, fun, low, print_event=True, full_return=False):
                 data["fun_args"] = list(args) + ([kwargs] if kwargs else [])
                 func_globals["__jid_event__"].fire_event(data, "new")
 
-                # Initialize a context for executing the method.
-                with salt.ext.tornado.stack_context.StackContext(
-                    self.functions.context_dict.clone
-                ):
```
meaksh, line 385: IIUC we can drop the `StackContext` completetly on newer tornado versions, but is it safe to drop it when running with older tornado versions (like the our vendored one)?

m-czernek: Good call, I haven't seen anything in the tests but now that you mention it, we can preserve the compatibility. WDYT about https://github.com/openSUSE/salt/pull/736/commits/69e8941baf3cf20cf5e527617864521cbaa6b1b4 ?

salt/modules/tls.py
```diff
@@ -1590,49 +1595,49 @@ def create_pkcs12(ca_name, CN, passphrase="", cacert_path=None, replace=False):
...
-    ):
+    p12_path = f"{cert_base_path()}/{ca_name}/certs/{CN}.p12"
+    ca_cert_path = f"{cert_base_path()}/{ca_name}/{ca_name}_ca_cert.crt"
+    cert_path = f"{cert_base_path()}/{ca_name}/certs/{CN}.crt"
+    priv_key_path = f"{cert_base_path()}/{ca_name}/certs/{CN}.key"
+
+    if not replace and os.path.exists(p12_path):
         return 'Certificate "{}" already exists'.format(CN)
 
+    # FIXME: error handling for load_pem_x509_certificate()
```
meaksh: Should we take care of this?

salt/modules/tls.py
```diff
@@ -120,6 +120,11 @@
 X509_EXT_ENABLED = True
 try:
     import OpenSSL
+    import cryptography.x509
+    import cryptography.exceptions
+
+    import cryptography.hazmat.primitives.serialization as cryptography_serialization
+    import cryptography.hazmat.primitives.serialization.pkcs12 as cryptography_pkcs12
```
meaksh, line 127: Hmm, this would add `python-cryptography` as new requirement for Salt, we would need to have this is as `Requires` in the spec, and also make sure the package is available for all the target OSes where we still submit the classic Salt package.

I see the package is available in SLE15SP3+, SLFO based OSes and also for RHEL8 based, so at a first glace we should be fine.

But do we want to make this move now, together with the changes at `x509` module for all OSes, or are we more conservatives and just add this change for newer OSes?

m-czernek: So there are the following commits:

- [`b10b2e9` (#736)](https://github.com/openSUSE/salt/pull/736/commits/b10b2e9e1c9281bc1e84da8f4666f1cf24bfbf89)
- [`19a606c` (#736)](https://github.com/openSUSE/salt/pull/736/commits/19a606c1c516d6d0f38c93a3dff7fbf2a4be748b)
- (and now my related fix) [`7bd54f9` (#736)](https://github.com/openSUSE/salt/pull/736/commits/7bd54f991fa6dd75feddc06eee84d77cb2033c1c)

Looking back at this, I see that `python3-pyOpenSSL` already requires it, which is why it's deployed on a lot of our test servers already, though I can't figure out what requires/recommends python3-pyOpenSSL. Do we not already require it through transitive dependencies?

If not, I'm fine either way honestly. We can revert the commits, or keep the commits in case we already transitively depend on it.

meaksh: Ok, for SLE, the `python3-cryptography` is pulled automatically when installing Salt because:
- `python3-salt` -- requires --> `python3-requests` -- requires --> `python3-urllib3` -- requires --> `python3-cryptography` & `python3-pyOpenSSL`

For RHEL8 based, that's not the case, as `python3-urlib3` does NOT requires `python3-cryptography` there, so we would need to explicitely add the `Requires` to the Salt spec file to be on the safe side.

For the Salt Bundle, we should be fine, as `python3-cryptography` is already included in the bundle.

So, I'd say to give this a try.

meaksh (approved):

salt/__init__.py
```diff
@@ -12,7 +12,7 @@
     )
     sys.stderr.flush()
 
-USE_VENDORED_TORNADO = True
+USE_VENDORED_TORNADO = True if sys.version_info < (3,12) else False
```
vzhestkov: Isn't it the same as?:
```suggestion
USE_VENDORED_TORNADO = sys.version_info < (3,12)
```

vzhestkov:
Hmm. I'm not sure about changing the imports from `salt.ext.tornado`, if I remember right `USE_VENDORED_TORNADO` exactly intended to change the behaviour on calling imports. If I remember this part of the code right, not fully sure now. Have you checked it if the behaviour really changing on renaming the imports? If not and `salt.ext.tornado` will keep working fine just with changing the value of `USE_VENDORED_TORNADO` I would prefer not to change them across all the files.

meaksh:
> Hmm. I'm not sure about changing the imports from `salt.ext.tornado`, if I remember right `USE_VENDORED_TORNADO` exactly intended to change the behaviour on calling imports. If I remember this part of the code right, not fully sure now. Have you checked it if the behaviour really changing on renaming the imports? If not and `salt.ext.tornado` will keep working fine just with changing the value of `USE_VENDORED_TORNADO` I would prefer not to change them across all the files.

Hmm, I did some tests now reverting the rename of imports and I got some errors + the mechanism `USE_VENDORED_TORNADO` not really working as expected: it keep using `salt.ext.tornado` instead of system's tornado when Python version is 3.13.

vzhestkov:
@meaksh, thanks for checking. I was not really sure about this point, sorry for the noise here.

I don't see anything else eye scratchy, but it's quite big to check. Let's try it :+1:

vzhestkov (approved):

agraul:
> Hmm. I'm not sure about changing the imports from `salt.ext.tornado`, if I remember right `USE_VENDORED_TORNADO` exactly intended to change the behaviour on calling imports. If I remember this part of the code right, not fully sure now. Have you checked it if the behaviour really changing on renaming the imports? If not and `salt.ext.tornado` will keep working fine just with changing the value of `USE_VENDORED_TORNADO` I would prefer not to change them across all the files.

`USE_VENDORED_TORNADO` needs a different implementation for newer Python interpreters, the import machinery work a bit differently now. https://github.com/openSUSE/salt/commit/0af5cc60aff2255c420372627d8ad7eee064f062 fixed it, but that didn't make it to PR.

PR #742, opened 2026-01-07
Extend fails to warnings until 2027

m-czernek:
### What does this PR do?

Extend warnings for another year, since we still fully support Salt 3006.


vzhestkov (approved):
LGTM

PR #747, opened 2026-02-09
Fix ansible.playbooks extra-vars quoting (bsc#1257831)

vzhestkov:
### What does this PR do?

Backport of https://github.com/saltstack/salt/pull/68787

Fixes the quoting issue on passing complex variables to the playbook execution which could cause issues on passing `--extgra-vars` on running `ansible-playbook`.

Tracks: https://github.com/SUSE/spacewalk/issues/29592

### Previous Behavior
Causing wrong args passing ot `ansible-playbook` in case of using unbalanced quoting and single quotes in general.

### New Behavior
Using proper quoting to pass the args, so no issues.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/modules/ansiblegate.py
+++ b/salt/modules/ansiblegate.py
@@ -18,6 +18,7 @@ import fnmatch
 import json
 import logging
 import os
+import shlex
 import subprocess
 import sys
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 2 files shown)


salt/modules/ansiblegate.py
```diff
@@ -367,9 +368,9 @@ def playbooks(
     if diff:
         command.append("--diff")
     if isinstance(extra_vars, dict):
-        command.append(f"--extra-vars='{json.dumps(extra_vars)}'")
+        command.append(f"--extra-vars={shlex.quote(json.dumps(extra_vars))}")
```
m-czernek, line 371: Do we want to execute this on Linux only? (per [0]). Given the feedback of the upstream port, this might come up anyways with upstream.

[0] https://docs.python.org/3/library/shlex.html#shlex.quote

vzhestkov: As far as I know this module never stated to be supported on windows. Even it's testing part contains:
```
pytestmark = [
    pytest.mark.skip_on_windows(reason="Not supported on Windows"),
]
```
We still have almost 0 feedback on this module and this particular issue was found while testing very specific complex variables passing to the module, not sure if anyone will try to do the same with windows.

m-czernek (approved):
Given that we maintain the module and it's unsupported on Windows, I'm OK with the change as is.

PR #750, opened 2026-03-12
Backport of the CVE-2026-31958 fix (bsc#1259554)

vzhestkov:
### What does this PR do?

Backport of https://github.com/tornadoweb/tornado/pull/3584 to vendored `tornado`.

The affected code is not used with the default configuration.
It makes sense to have the fix to avoid shipping the code affected with CVE.

Tracks: https://github.com/SUSE/spacewalk/issues/29985

### Previous Behavior
The affected code is not used with the default deployment, so there should be no any changes in behaviour.

### New Behavior
The affected code is not used with the default deployment, so there should be no any changes in behaviour.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/ext/tornado/httputil.py
+++ b/salt/ext/tornado/httputil.py
@@ -71,6 +71,10 @@ except ImportError:
 # To be used with str.strip() and related methods.
 HTTP_WHITESPACE = " \t"
 
+# Roughly the inverse of RequestHandler._VALID_HEADER_CHARS, but permits
+# chars greater than \xFF (which may appear after decoding utf8).
+_FORBIDDEN_HEADER_CHARS_RE = re.compile(r"[\x00-\x08\x0A-\x1F\x7F]")
+
 # RFC 7230 section 3.5: a recipient MAY recognize a single LF as a line
 # terminator and ignore any preceding CR.
 _CRLF_RE = re.compile(r"\r?\n")
@@ -155,6 +159,8 @@ class HTTPHeaders(MutableMapping):
     def add(self, name, value):
         # type: (str, str) -> None
         """Adds a new value for the given key."""
+        if _FORBIDDEN_HEADER_CHARS_RE.search(value):
+            raise HTTPInputError("Invalid header value %r" % value)
         norm_name = _normalized_headers[name]
         self._last_key = norm_name
         if norm_name in self:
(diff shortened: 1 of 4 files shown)


salt/ext/tornado/httputil.py
```diff
@@ -187,13 +193,31 @@ def parse_line(self, line):
...
+        if not line:
+            # Empty line, or the final CRLF of a header block.
+            return
+        if line[0] in HTTP_WHITESPACE:
+            # continuation of a multi-line header
+            # TODO(7.0): Remove support for line folding.
+            if self._last_key is None:
+                raise HTTPInputError("first header line cannot start with whitespace")
             # continuation of a multi-line header
             new_part = " " + line.lstrip(HTTP_WHITESPACE)
```
m-czernek: Just a note here, this is now `strip` instead of `lstrip`.

salt/ext/tornado/httputil.py
```diff
@@ -753,7 +777,84 @@ def _int_or_none(val):
     return int(val)
 
 
-def parse_body_arguments(content_type, body, arguments, files, headers=None):
+class ParseMultipartConfig:
```
m-czernek, line 779: Are we missing the `__eq__` method? Something like:

```
def __eq__(self, other):
        if not isinstance(other, ParseMultipartConfig):
            return NotImplemented
        return (
            self.enabled == other.enabled and
            self.max_parts == other.max_parts and
            self.max_part_header_size == other.max_part_header_size
        )
```

m-czernek (approved):
This LGTM, leaving just a few comments of what I noticed during the review.

salt/ext/tornado/httputil.py
vzhestkov: Fixed, thanks for catching :+1: . Seems missed it on rebasing and besides of it there was a doubled comment close to this line.

salt/ext/tornado/httputil.py
vzhestkov: I don't think it's needed. In the existing implementation the default config is only used. It could be only used if someone develop something exposing the server which should parse multipart valuses sent with the form, what shouldn't be the case also.

PR #753, opened 2026-04-02
Read full URI from ldap pillar config

agraul:
A full URI allows the user to set a scheme for the ldap connection and enables tls. This is a workaround for the ldap execution module, which is used by the ldap pillar module.

While the ldap auth module supports both "tls" and "starttls" modes, the ldap execution module only supports "starttls", which it calls "tls".

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1254900


ycedres (approved):
LGTM!

Issue #740, opened 2025-11-28
[BUG] gpgkey is filtered out by the normalize pkgrepo function for suse


darix:
the documentation says you can pass the path for the gpgkey but all unknown fields are dropped

https://github.com/openSUSE/salt/blob/openSUSE/release/3006.0/salt/states/pkgrepo.py#L754-L776

please allow all fields from the repo file format to pass through

PR #769, opened 2026-06-29
Repair gpgkey/refresh handling in pkgrepo.managed

tacerus:
Hi,

I wanted to submit this upstream, but I notice some dependent commits are missing from 3006.x and I could not find any pending PRs for them.

### What does this PR do?

These were so implicitly ignored so far.
Zypper command line does not support modification of "gpgkey", hence extend functions to handle it via ConfigParser and expand the tests respectively.
The logic from the previous change "Fix zypper repository reconfiguration" is reverted in favor of including the missing attributes in _get_repo_info(). The gpgkey option itself was not included there, as returning "None" in case of no gpgkey might be confusing, given zypp implying a gpgkey if none is set if the baseurl matches a hardcoded list of mirrors.

Fixes:
- https://github.com/saltstack/salt/issues/60520
- https://github.com/openSUSE/salt/issues/740

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1188,17 +1188,40 @@ def list_repo_pkgs(*args, **kwargs):
         return byrepo_ret
 
 
+def _get_repos_directory(root=None):
+    return os.path.join(root, os.path.relpath(REPOS, os.path.sep)) if root else REPOS
+
+
+def _get_configured_repo(repo, root=None):
+    repos = _get_repos_directory(root=root)
+
+    for fname in os.listdir(repos):
+        if not fname.endswith(".repo"):
+            continue
+
+        repo_file = os.path.join(repos, fname)
+
+        repo_cfg = configparser.ConfigParser()
+        repo_cfg.read(repo_file)
+
+        if repo in repo_cfg:
+            # return whole config for processing as there might be multiple repository sections in one file
+            return {repo_file: repo_cfg}
+
+    return {}
+
+
 def _get_configured_repos(root=None):
     """
     Get all the info about repositories from the configurations.
     """
 
-    repos = os.path.join(root, os.path.relpath(REPOS, os.path.sep)) if root else REPOS
+    repos = _get_repos_directory(root=root)
     repos_cfg = configparser.ConfigParser()
     if os.path.exists(repos):
         repos_cfg.read(
             [
-                repos + "/" + fname
+                os.path.join(repos, fname)
                 for fname in os.listdir(repos)
                 if fname.endswith(".repo")
             ]
@@ -1219,10 +1242,14 @@ def _get_repo_info(alias, repos_cfg=None, root=None):
         )
         meta["alias"] = alias
         for key, val in meta.items():
-            if val in ["0", "1"]:
+            if key == "priority":
+                meta[key] = int(meta[key])
+            elif val in ["0", "1"]:
                 meta[key] = int(meta[key]) == 1
             elif val == "NONE":
                 meta[key] = None
+        if "priority" not in meta:
+            meta["priority"] = DEFAULT_PRIORITY
         return meta
     except (ValueError, configparser.NoSectionError):
         return {}
@@ -1444,6 +1471,18 @@ def mod_repo(repo, **kwargs):
     if cmd_opt:
         cmd_opt = global_cmd_opt + ["mr"] + cmd_opt + [repo]
         __zypper__(root=root).refreshable.xml.call(*cmd_opt)
+
+    if "gpgkey" in kwargs:
+        # TODO: factor file based configuration to helper function
+        repo_map = _get_configured_repo(repo, root=root)
+        if repo_map:
+            repo_file, repo_file_cfg = tuple(repo_map.items())[0]
+            repo_cfg = repo_file_cfg[repo]
+
+            if repo_cfg.get("gpgkey") != kwargs["gpgkey"]:
+                repo_cfg["gpgkey"] = kwargs["gpgkey"]
+                with salt.utils.files.fopen(repo_file, "w") as _fp:
+                    repo_file_cfg.write(_fp)
 
     comment = None
     if call_refresh:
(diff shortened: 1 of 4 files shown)


tacerus:
while amending the tests, I also noticed that "comments" is getting ignored - I wonder how useful that feature is so I left it as a separate issue.

salt/modules/zypperpkg.py
```diff
@@ -1188,21 +1188,40 @@ def _is_match(pkgname):
...
+                repo_file = os.path.join(repos, fname)
+
+                repo_cfg = configparser.ConfigParser()
+                repo_cfg.read(repo_file)
+
+                if repo in repo_cfg:
+                    # return whole config for processing as there might be multiple repository sections in one file
+                    return {repo_file: repo_cfg}
+
+            return {}
```
m-czernek: We now have one helper method that has two return types, depending on a parameter. Please stick to one return type per function. This should probably be its own helper function.

salt/modules/zypperpkg.py
```diff
@@ -1223,6 +1242,10 @@ def _get_repo_info(alias, repos_cfg=None, root=None):
                 meta[key] = int(meta[key]) == 1
             elif val == "NONE":
                 meta[key] = None
+            elif key == "priority":
+                meta[key] = int(meta[key])
```
m-czernek: Seems like here, if `priority` equals to `0` or `1`, it will be caught by this branch:

```
            if val in ["0", "1"]:
                meta[key] = int(meta[key]) == 1
```

And turned to `True` or `False`. This check should probably be the first in the sequence, not the last.

salt/states/pkgrepo.py
```diff
@@ -792,7 +795,7 @@ def absent(name, **kwargs):
     return ret
 
 
-def _normalize_repo(repo):
+def _normalize_repo(repo, for_mod=False):
```
m-czernek, line 798: Can you explain to me what `for_mod` means here, and why it drops `humanname` for SUSE OSes?

salt/modules/zypperpkg.py
tacerus: Hi, thanks for reviewing. Good point, I refactored it to a dedicated `_get_configured_repo()` (singular) function.

salt/modules/zypperpkg.py
tacerus: Good catch, I adjusted it.

salt/states/pkgrepo.py
tacerus: It alters the behavior if the function is called for modifications, but I'm afraid, I am also confused about why this is needed. If I include the `humanname` mapping during modify operations, the tests will fail with things like

```
E                 Left contains 1 more item:
E                 {'humanname': {'new': 'Salt modified Backports',
E                                'old': 'openSUSE Backports for SLE 15 SP4'}}
E                 Right contains 1 more item:
E                 {'name': {'new': 'Salt modified Backports',
E                           'old': 'openSUSE Backports for SLE 15 SP4'}}
```

There might be another place this `humanname` => `name` mapping needs to be added in, but I failed to figure it out.

Would you have some better idea?

m-czernek: Ok, so after looking into this:

The problem here is IMO adding the `_normalize_repo` call in the `managed` function. This hides a couple of problems, because it removes arguments that do not exist in it.

The only translation you need is `autorefresh` to `refresh`, and we can drop `comments` since they don't actually appear in the repo files.

I would recommend this patch (feel free to modify it):

```diff
diff --git i/salt/states/pkgrepo.py w/salt/states/pkgrepo.py
index af714a24a2..7a6419f3cc 100644
--- i/salt/states/pkgrepo.py
+++ w/salt/states/pkgrepo.py
@@ -484,8 +484,9 @@ def managed(name, ppa=None, copr=None, aptkey=True, **kwargs):
         sanitizedkwargs = kwargs
 
         if __grains__["os_family"] == "Suse":
-            pre = _normalize_repo(pre, for_mod=True)
-            sanitizedkwargs = _normalize_repo(sanitizedkwargs, for_mod=True)
+            _suse_repo_keys_to_mod_repo(pre)
+            # zypper cannot store comments in a .repo file
+            sanitizedkwargs.pop("comments", None)
 
     if pre:
         # 22412: Remove file attribute in case same repo is set up multiple times but with different files
@@ -609,7 +610,7 @@ def managed(name, ppa=None, copr=None, aptkey=True, **kwargs):
     try:
         post = __salt__["pkg.get_repo"](repo=repo, **kwargs)
         if __grains__["os_family"] == "Suse":
-            post = _normalize_repo(post, for_mod=True)
+            _suse_repo_keys_to_mod_repo(post)
         if pre:
             for kwarg in sanitizedkwargs:
                 if post.get(kwarg) != pre.get(kwarg):
@@ -795,7 +796,16 @@ def absent(name, **kwargs):
     return ret
 
 
-def _normalize_repo(repo, for_mod=False):
+def _suse_repo_keys_to_mod_repo(repo):
+    """Rename in place the keys reported by `pkg.get_repo()` that differ from
+    the ones `pkg.mod_repo()` accepts, so both can be compared.
+    """
+    if "autorefresh" in repo:
+        repo["refresh"] = repo.pop("autorefresh")
+    return repo
+
+
+def _normalize_repo(repo):
     """Normalize the get_repo information"""
     # `pkg.get_repo()` specific virtual module implementation is
     # parsing the information directly from the repository
@@ -815,8 +825,6 @@ def _normalize_repo(repo, for_mod=False):
         "keepackages": "cache",
         "baseurl": "url",
     }
-    if not for_mod:
-        suse["name"] = "humanname"
 
     translator = {
         "Suse": suse,
```

m-czernek: One more thing that is tangentially related - after the tests run, the repo files stay defined. This is not a huge problem (non-blocking for the PR), but if you can remove them at the end of the test, that'd be helpful.

PR #779, opened 2026-08-27
Fix test_tcp for pytest >=8

m-czernek:
### What does this PR do?

When executing this test with pytest>=8, it fails with the following error:

```
04:28:49  ERROR tests/unit/transport/test_tcp.py::AsyncTestCase - AttributeError: 'AsyncTestCase' object has no attribute 'runTest'
04:28:49  ERROR tests/unit/transport/test_tcp.py::AsyncPubServerTest - AttributeError: 'AsyncPubServerTest' object has no attribute 'runTest'
```

This commit fixes the failure. This is a continuation of https://github.com/openSUSE/salt/pull/774 

I also realized that while upstream is migrating the tests to pytest (which fixes the issue), we might want to fix these upstream as well in case the migration takes a long time, so I created https://github.com/saltstack/salt/pull/70172 which contains all the relevant fixes in https://github.com/openSUSE/salt/pull/774 and this PR. 

### Previous Behavior
Tests errors out

### New Behavior
Test can execute

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/tests/unit/transport/test_tcp.py
+++ b/tests/unit/transport/test_tcp.py
@@ -17,7 +17,7 @@ import tornado.gen
 import tornado.ioloop
 import salt.utils.platform
 import salt.utils.process
-from tornado.testing import AsyncTestCase
+import tornado.testing
 from tests.support.mixins import AdaptedConfigurationTestCaseMixin
 from tests.unit.transport.mixins import run_loop_in_thread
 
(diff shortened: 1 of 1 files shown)


vzhestkov (approved):
Looks good to me, but just curious if there any difference of behaviour between:
```
import tornado.testing
```
and
```
from tornado.testing import AsyncTestCase
```
when only `tornado.testing.AsyncTestCase` is used later anyway?

m-czernek:
@vzhestkov yes, when importing `AsyncTestCase` directly, the base class is evaluated for the presence of the `runTest` attribute. You can see this from the error itself:

```
ERROR tests/unit/transport/test_tcp.py::AsyncTestCase - AttributeError: 'AsyncTestCase' object has no attribute 'runTest'
```

PR #780, opened 2026-09-01
Support the oscap options that select rules and content in xccdf_eval

jniedergang:
### What does this PR do?

`openscap.xccdf_eval` rebuilds the oscap command line from a fixed set of keyword arguments, so
anything else a caller asks for is dropped without a word. SUSE Multi-Linux Manager exposes an oscap
arguments field that ends up here, which means a user asking to skip a rule gets a compliance report
where that rule was evaluated anyway.

This adds the options that select which rules are evaluated and where the content comes from:
`skip_rule`, `reference`, `cpe`, `local_files`, `datastream_id`, `xccdf_id` and `benchmark_id`.
`rule` and `skip_rule` now accept either a single string or a list, since oscap accepts those options
more than once.

The options that write result files (`results`, `results-arf`, `report`, `stig-viewer`,
`thin-results`) are deliberately left out: the caller owns those paths and already passes them
explicitly.

References the SUSE Multi-Linux Manager side of the same problem:
- issue: https://github.com/uyuni-project/uyuni/issues/12538
- pull request: https://github.com/uyuni-project/uyuni/pull/12539

### Previous Behavior

```
salt '*' openscap.xccdf_eval /path/to/ds.xml profile=Default skip_rule=rule-one
```

runs `oscap xccdf eval --profile Default /path/to/ds.xml`: the rule is evaluated.

### New Behavior

The same call runs `oscap xccdf eval --profile Default --skip-rule rule-one /path/to/ds.xml`, and
`skip_rule=["rule-one", "rule-two"]` produces one `--skip-rule` per rule.

- [x] Docs (the function docstring documents every new argument)
- [x] Changelog (`changelog/780.added.md`)
- [x] Tests written/updated (`tests/unit/modules/test_openscap.py`: 14 tests pass, including the 11
      pre-existing ones; black 22.6.0 and isort 5.12.0 clean on the added lines)

### Commits signed with GPG?

Yes (SSH signature, shown as Verified by GitHub).

### DIFF
--- a/changelog/780.added.md
+++ b/changelog/780.added.md
@@ -1,0 +1,1 @@
+Added the oscap options selecting rules and content to `openscap.xccdf_eval`: `skip_rule`, `reference`, `cpe`, `local_files`, `datastream_id`, `xccdf_id` and `benchmark_id`. `rule` and `skip_rule` now accept a list, since oscap accepts those options more than once.
--- a/salt/modules/openscap.py
+++ b/salt/modules/openscap.py
@@ -56,6 +56,18 @@ _OSCAP_EXIT_CODES_MAP = {
 }
 
 
+def _append_repeatable(cmd_opts, option, value):
+    """
+    Append one occurrence of a repeatable oscap option per value.
+
+    The value is either a single string or a list of strings.
+    """
+    values = [value] if isinstance(value, str) else value
+    for item in values:
+        cmd_opts.append(option)
+        cmd_opts.append(item)
+
+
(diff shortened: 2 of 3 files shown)


mcalmer:
We have re-implemented the OpenSCAP UI and backend already. This seems to apply changes to the legacy implementation.
I assume, we will drop this code soon.
Check the new feature:
https://www.uyuni-project.org/uyuni-docs/en/uyuni/administration/openscap-enhanced-beta.html

